PR #14, opened 2020-12-10
Add Dragino Package

PaulVMo:
Added make profile for the mips_24c platform including gateways such as the Dragino LPS8 which this was tested on


madninja:
Hmm.. so the Dragino is not mipsel? Interesting

PaulVMo:
@madninja I'm learning as I go here. The Dragino uses a Atheros AR9331 which is MIPS 24K so I gave it a shot with mips instead of mipsel based on what I found on the Rust supported platforms. I could be wrong - it builds and runs though :)

madninja (approved):

madninja:
Thanks @PaulVMo !

PR #17, opened 2020-12-15
First attempt at RasPi support

shawaj:
This is a first draft commit and a work in progress PR.

It cross compiles fine using `cross build` but I am having some issues with getting cargo-deb to work properly for packaging the deb file. Perhaps might be a better idea to package the .deb file manually like is done with the .ipk

Some other things to note:
* For cargo-deb to work I had to install `cmake` and `gcc-arm-linux-gnueabihf` and `rustup target add armv7-unknown-linux-gnueabihf`
* The ci.yml for github actions is a mess - need to fix

Let me know if you spot any issues or have any suggestions.


shawaj:
Just for reference, the error I am getting with cargo deb is as follows:

```
          /usr/bin/ld: /home/GitHub/gateway-rs/target/arm-unknown-linux-gnueabihf/release/deps/helium_gateway-0a4717c34b6d41e2.helium_gateway.1o693kuq-cgu.0.rcgu.o: relocations in generic ELF (EM: 40)
          /usr/bin/ld: /home/GitHub/gateway-rs/target/arm-unknown-linux-gnueabihf/release/deps/helium_gateway-0a4717c34b6d41e2.helium_gateway.1o693kuq-cgu.0.rcgu.o: error adding symbols: file in wrong format
          collect2: error: ld returned 1 exit status
```

shawaj:
Cargo-deb is basically ignoring the `.config/config.toml` due to the file extension...

As per the [docs here](https://doc.rust-lang.org/cargo/reference/config.html) the preferred config is `.config/config.toml` instead of `.config/config` but due to the code here:

https://github.com/mmstick/cargo-deb/blob/b5e4f398545a42ecbfdae7175f47c09cde103655/src/config.rs#L38

It is only looking at `.config/config`

See issue on cargo-deb here https://github.com/mmstick/cargo-deb/issues/157

shawaj:
@madninja the fix in cargo-deb is in and so in future versions it should be ok with a config.toml.

Are you happy with the config being `.config/config` without the toml extension for now, until the new cargo-deb package is in the repos? Just want to make sure it doesn't break anything.

madninja:
@shawaj ok, I think we're ready to look at how this gets merged, if you're still interested? 

Does the config vs config.toml work for the non cargo-deb platforms?

madninja:
Thanks for your efforts here @shawaj! I'm going to close this in favor of: https://github.com/helium/gateway-rs/pull/37 which takes your initial constructs and fits it more into the cargo-make approach while making packaging consistent across ipk/deb targets

shawaj:
Thanks @madninja 

Cc @ryanteck

Issue #30, opened 2021-04-07
RAK7258 error


argonfatl:
Not sure why but i have that 
way[1531]:  INFO routing error: Service(Rpc(Status { code: Unknown, message: "h2 protocol error: protocol error: received frame when stream half-closed" })), module: router
Wed Apr  7 08:26:32 2021 daemon.info helium_gateway[1531]:  INFO selected gateway, uri: http://18.216.219.228:8080/, public_key: 112AY1cJ9quokGVZL8uVQ9Jih1cHsji2VVoKApf2fVPBseHcRsg, module: router
Wed Apr  7 08:26:33 2021 daemon.info helium_gateway[1531]:  INFO updated routing to height 793801, module: router

lthiery:
@argonfatl thanks for the report! We're trying to resolve it as it appears to affect downlinks (JoinAccepts included)

argonfatl:
Fri Apr  9 12:40:37 2021 daemon.info helium_gateway[1531]:  WARN router returned invalid height 795354 while at 795354, module: router
Fri Apr  9 12:40:37 2021 daemon.info helium_gateway[1531]:  INFO updated routing to height 795354, module: router
Fri Apr  9 12:44:40 2021 daemon.info helium_gateway[1531]:  INFO no update found, module: updater
Fri Apr  9 12:54:40 2021 daemon.info helium_gateway[1531]:  INFO no update found, module: updater
Fri Apr  9 13:04:40 2021 daemon.info helium_gateway[1531]:  INFO no update found, module: updater

lthiery:
@argonfatl I believe this issue was resolved with yesterday's Router release, so I think I'll close this issue.

The logs you are showing aren't particular concerning? The warning is perhaps a little overzealous?

lthiery:
For extra context, here was the bug fix
https://github.com/helium/router/pull/277

argonfatl:
after Rak firmware update, everything looks like  works good

PR #31, opened 2021-04-09
Height warning

lthiery:
Is it a concern if local height and service height are equal?

### DIFF
--- a/src/router/mod.rs
+++ b/src/router/mod.rs
@@ -106,7 +106,7 @@ impl Router {
     }
 
     fn handle_routing_update(&mut self, logger: &Logger, routing_response: &RoutingResponse) {
-        if routing_response.height <= self.routing_height {
+        if routing_response.height < self.routing_height {
             warn!(
                 logger,
                 "router returned invalid height {:?} while at {:?}",


src/router/mod.rs
```diff
@@ -106,7 +106,7 @@ impl Router {
     }
 
     fn handle_routing_update(&mut self, logger: &Logger, routing_response: &RoutingResponse) {
-        if routing_response.height <= self.routing_height {
+        if routing_response.height < self.routing_height {
```
madninja, line 109: mostly because you've already processed the state at that block. Are you seeing issues where it's returning the same block again? Seems odd

madninja, line 109: Because the client would already have processed routing info at that height, so something fishy is going on? Are you seeing this happen?

lthiery: It was reported [here](https://github.com/helium/gateway-rs/issues/30#issuecomment-816880282) in an unrelated issue.

madninja: That's an actual problem with the gateway service then.. and the warning is correct. It's not terminal but it does indicate the gateway service is doing something not quite right

lthiery: Appropriate warning then!

PR #34, opened 2021-04-17
add region setting in README

said026:
I've added the region setting in the README file (somehow with my Dragino LPS8 I had to edit it manually, as I am in EU868 region), not sure if I listed all possible regions in my edit : 
`US915| EU868 | EU433 | CN470 | CN779 | AU915 | AS923 | KR920 | IN865`

I don't know if it is possible to "autodetect" the region from the platform settings directly, this might be useful :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,11 @@ If your [supported LoRa gateway](#supported-platforms) did not come with helium-
    ```
    **NOTE**: Some platform have custom package installation requirements. Refer to the developer instructions for that platform on how to install a package.
 
-   The default region of the gateway is `US915`, if your region is different you can set the right one in `/etc/helium_gateway/default.toml`. Possible values are : `US915| EU868 | EU433 | CN470 | CN779 | AU915 | AS923 | KR920 | IN865`, after updating the value you need to restart the service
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,11 @@ If your [supported LoRa gateway](#supported-platforms) did not come with helium-
    ```
    **NOTE**: Some platform have custom package installation requirements. Refer to the developer instructions for that platform on how to install a package.
 
+   The default region of the gateway is `US915`, if your region is different you can set the right one in `/etc/helium_gateway/default.toml`. Possible values are : `US915| EU868 | EU433 | CN470 | CN779 | AU915 | AS923 | KR920 | IN865`, after updating the value you need to restart the service
```
madninja, line 57: You should never change default.toml. Make changes only to settings.toml for local overrides

madninja (changes requested):

madninja:
> I don't know if it is possible to "autodetect" the region from the platform settings directly, this might be useful :)

If there was such a thing as a  "platform setting" where we could infer the region that would be cool, but it seems highly intertwined with the lora packet forwarder settings on the same device which are quite different per platform.

README.md
said026: Thanks for the hint, just updated the Readme file 👍

madninja (approved):

PR #38, opened 2021-04-21
gitignore: settings.toml and key.bin

lthiery:


lthiery:
These files are likely added during development for testing. This prevents accidental commits.

madninja (approved):

PR #56, opened 2021-06-18
adding support for ResIOT gateways

StevenResiot:


package/resiot/control/control
```diff
@@ -0,0 +1,6 @@
+Package: helium_gateway
+Version: ${CARGO_MAKE_CRATE_VERSION}
+Architecture: armv7ahf-vfp
```
madninja, line 3: This should probably be ${CARGO_MAKE_PROFILE} since it now appears to point at slightly different architecture?

StevenResiot: Had to pick one from our opkg architecture list as the replaced variable does not appear in the default list. I assumed editing it here was faster than reflashing all the gateways or manually editing opkg configuration settings on the gateways before installing

madninja: ugh.. you'd think that the architecture would match the build arch target.. oh well. if it works for ya

madninja:
If you could rebase your PR on main to fix up the ci flow and readme?

Makefile.toml
```diff
@@ -101,7 +106,7 @@ script = '''
 	cd ${CARGO_MAKE_WORKING_DIRECTORY}
 
 	cd ${PKG_BUILD_DIR}
-	if [ "true" = "${IPK_USE_AR}-unset}" ]; then
+	if [ "true" = "${IPK_USE_AR}" ]; then
 		rm ../${PKG_NAME}
```
madninja, line 110:
```suggestion
		rm -f ../${PKG_NAME}
```

package/resiot/control/prerm
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+/etc/init.d/helium_gateway stop
```
madninja, line 3: remove this line.. auto upgrades when the system tries to stop during the upgrade

madninja (changes requested):

madninja (approved):

PR #62, opened 2021-07-06
work around cargo-make env issue

madninja:


PaulVMo:
I tested this branch on both a Dragino LPS8 and RAK 7258 by building and installing manually. The install was successfully. This fix appears to resolve the issue of the hanging install for Draginos on the latest build.

PR #64, opened 2021-07-07
Add raspi64 profile (pi3/4 using 64 bit raspi os)

fishpepper:
This PR adds the support to build for a raspberry pi 3 or 4 using the 64 bit raspberry OS.
I am not sure if raspi64 is the best name choice, maybe raspi34_64 would be better?

I plan on using this profile on a pi3 with the SeeedStudio WM1302 Raspberry Pi Hat once my hardware arrives.


madninja:
you can run the existing raspi234 package on an 64bit Raspberry OS no?

fishpepper:
In principal yes, but the default Raspberry Pi OS in 64 bit does not come with the 32 bit compatibility... There are armhf libs installed.
Building gateway-rs natively for aarch64 seems to work fine. I am still waiting for my WM1302 to arrive in order to properly test all functionality. The module is finally on its way after all those covid caused chip delays...

fishpepper:
After some days testing it i can confirm the arm64 build works as expected and forwards data from my LHT65 sensor to the staging console.

madninja:
> After some days testing it i can confirm the arm64 build works as expected and forwards data from my LHT65 sensor to the staging console.

Awesome! Would you mind rebasing your branch so and resolving the conflict?

fishpepper:
there you go :)

madninja:
> there you go :)

Last request: Can you add this to the build matrix, please?

fishpepper:
Sure! II missed the .github dir during my grep to find places that needed the addition of raspi64 ;)

madninja:
> Sure! II missed the .github dir during my grep to find places that needed the addition of raspi64 ;)

Sorry one more.. The 64 bit designator for x86 is x86_64 and raspi012 and raspi34 already have numbers right after the name. 
To be consistent could you rename it to `raspi_64`.

PR #74, opened 2021-07-16
improve logs in gateway for received frames

lthiery:
Previously, every single UDP frame was printed at `info` level with the explanation "ignoring raw packet". Whether "ignoring" is good or bad isn't really clear to the user, so the new explanation is "GWMP frame received" and the log-level is bumped down to `debug`.

However, `PushData` frames are specially handled and continue to be printed at the `info` level. 

Whereas before they printed as so:
```
INFO ignoring raw packet PushData(Packet { random_token: 27289, gateway_mac: MacAddress { bytes: [0, 0, 0, 0, 4, 3, 2, 1] }, data: Data { rxpk: Some([V1(RxPkV1 { chan: 0, codr: _4_5, data: [0, 1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 3, 4, 5, 6, 7, 8, 109, 176, 243, 16, 109, 161], datr: DataRate(SF10, BW125), freq: 904.7, lsnr: 5.5, modu: LORA, rfch: 0, rssi: -112, rssis: None, size: 23, stat: OK, tmst: 454 })]), stat: None } }), module: gateway
```

Now, they display slightly more succinctly:
```
INFO uplink from MacAddress(00:00:00:00:04:03:01), 904.3 MHz, DataRate(SF10, BW125), rssic = -112, snr = 5.5,len = 23, module: gateway
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ xxhash-c = "0.8"
 xorf = "0.7"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-semtech-udp = { version = "^0.5.2", default-features=false, features=["server"] }
+semtech-udp = { version = "^0.6", default-features=false, features=["server"] }
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.2.1"}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ xxhash-c = "0.8"
 xorf = "0.7"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-semtech-udp = { version = "0", default-features=false, features=["server"] }
+semtech-udp = { version = "^0.5.2", default-features=false, features=["server"] }
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.2.1"}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ xxhash-c = "0.8"
 xorf = "0.7"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-semtech-udp = { version = ">=0.6", default-features=false, features=["server"] }
+semtech-udp = { version = ">=0.6,<1", default-features=false, features=["server"] }
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.2.1"}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,7 +45,7 @@ xxhash-c = "0.8"
 xorf = "0.7"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-semtech-udp = { version = "^0.6", default-features=false, features=["server"] }
+semtech-udp = { version = ">=0.6", default-features=false, features=["server"] }
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.2.1"}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -89,12 +89,7 @@ impl Gateway {
                     if let Some(rxpks) = packet.data.rxpk {
                         let mac = packet.gateway_mac;
                         for rxpk in rxpks {
-                            info!(
-                                logger,
-                                "uplink {}, from {}",
-                                rxpk,
-                                mac
-                            )
+                            info!(logger, "uplink {}, from {}", rxpk, mac)
                         }
                     }
                 }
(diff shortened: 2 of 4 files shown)


src/gateway.rs
```diff
@@ -84,9 +84,30 @@ impl Gateway {
...
+            Event::RawPacket(raw) => match raw {
+                semtech_udp::Up::PushData(packet) => {
+                    if let Some(rxpks) = packet.data.rxpk {
+                        let mac = packet.gateway_mac;
+                        for rxpk in rxpks {
+                            info!(
+                                logger,
+                                "uplink from {}, {} MHz, {:?}, {}, snr = {},len = {}",
+                                mac,
+                                rxpk.get_frequency(),
```
madninja, line 96: could you make this the rxpk Display trait?

madninja (commented):
One implementation detail and:

Why is this trying to say "uplink from" when it's not going to get sent upstream at all?

src/gateway.rs
lthiery: Yes, that sounds pretty reasonable. I'll make that change in `semtech-udp`

lthiery:
Implemented display for both uplinks and downlinks:
```
 INFO uplink @1555844454 ms, 904.70 MHz, DataRate(SF10, BW125), rssic: -112, snr: 5.5, len: 23, from MacAddress(01:02:03:04:05:06:08), module: gateway
 INFO routing packet to: http://52.8.80.146:8080/, module: router
 INFO rx1 downlink @1560844454 ms, 925.70 MHz, DataRate(SF10, BW500), len: 17 via MacAddress(01:02:03:04:05:06:08), module: gateway
 INFO uplink @1561945960 ms, 904.70 MHz, DataRate(SF10, BW125), rssic: -112, snr: 5.5, len: 23, from MacAddress(01:02:03:04:05:06:08), module: gateway
 INFO routing packet to: http://52.8.80.146:8080/, module: router
 INFO rx1 downlink @1566945960 ms, 925.70 MHz, DataRate(SF10, BW500), len: 17 via MacAddress(01:02:03:04:05:06:08), module: gateway
 INFO uplink @1568046329 ms, 904.10 MHz, DataRate(SF10, BW125), rssic: -112, snr: 5.5, len: 23, from MacAddress(01:02:03:04:05:06:08), module: gateway
 INFO routing packet to: http://52.8.80.146:8080/, module: router
 INFO rx1 downlink @1573046329 ms, 923.90 MHz, DataRate(SF10, BW500), len: 17 via MacAddress(01:02:03:04:05:06:08), module: gateway
 ```

I also did some refactoring of an "if error return, else unwrap". I can revert that if you like, but I don't know idiomatic the "if error return, else unwrap" approach is. If nesting is a visual problem, I can break out the subsection as a function?

File: Cargo.toml [hunk 2]
```diff
@@ -45,7 +45,7 @@ xxhash-c = "0.8"
 xorf = "0.7"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-semtech-udp = { version = "0", default-features=false, features=["server"] }
+semtech-udp = { version = "^0.5.2", default-features=false, features=["server"] }
```
madninja, line 48: This should not be needed? version 0 with a cargo update would pick up the newer version

src/gateway.rs
```diff
@@ -100,43 +108,55 @@ impl Gateway {
...
+                                "rx2 downlink {} via {}",
+                                txpk,
+                                downlink_rx2.get_destination_mac()
+                            );
+                            downlink_rx2.set_packet(txpk);
+                            match downlink_rx2
+                                .dispatch(Some(Duration::from_secs(DOWNLINK_TIMEOUT_SECS)))
+                                .await
+                            {
+                                Err(SemtechError::Ack(tx_ack::Error::NONE)) => Ok(()),
```
madninja, line 141: wut.. what is an Error::NONE but an Ok in disguise?

madninja (commented):
* I think the Error::NONE is confusing the logic here
* Also fix the clippy warning

Cargo.toml
lthiery: I added a new API for getting the MAC of a downlink in `0.5.2` and it doesn't exist in `0.5.1`. I think this is better form than depending on the lock?

src/gateway.rs
lthiery: This was your code! :P

but yes, I can try to fix this upstream but let's do in another PR? I think this is an artifact of lazy GWMP deserializing

madninja: It was indeed, but since I have you :-)

lthiery: oh hey i think its fixed already kind of

madninja: uhh how? So many Error::NONE to remove?

lthiery: Dispatch returns error by asking `ack.get_error()`

https://github.com/helium/semtech-udp/blob/master/src/server_runtime/mod.rs#L107

which only returns an error if there it's not `NO_ERROR`

https://github.com/helium/semtech-udp/blob/master/src/packet/tx_ack.rs#L30

madninja: nice! Want to adjust this PR?

lthiery: I'm looking at it... the NO_ERROR still gets kicked back and would require handling at this level.

This is the best I can do without improving the deserializer:
```
Ok(()) | Err(SemtechError::Ack(tx_ack::Error::NONE)) => Ok(()),
```

lthiery: I'm looking at deserializer improvements to see if i can swing it

lthiery: [Changes made in `semtech-udp`](https://github.com/helium/semtech-udp/pull/33) to improve the `tx_ack::Error` type. Release `0.6.0` includes it

lthiery:
There’s nothing I’m blocking on for this is there? @madninja

madninja:
> There’s nothing I’m blocking on for this is there? @madninja

Does the `~0.6` spec now mean it won't pick up 0.7? Any way to spec out a 0 major and any minor version >= 6?

lthiery:
> > There’s nothing I’m blocking on for this is there? @madninja
> 
> Does the `~0.6` spec now mean it won't pick up 0.7? Any way to spec out a 0 major and any minor version >= 6?

I think this is the way to do it. It's constrained `0.6< semtech_udp_version < 1` now.

madninja (approved):

Issue #80, opened 2021-08-05
Add support for Bobcat light gateway series


JackeyTY:
Hi,

We'd also want to add support for Bobcat light gateway series. Target: arm-openwrt-linux-muslgnueabi.

We tried to install klkgw ipk and it failed with;
root@OpenWrt-G220:~# opkg install helium-gateway-v1.0.0-alpha.13-klkgw.ipk 
Unknown package 'helium_gateway'.
Collected errors:
 * pkg_hash_fetch_best_installation_candidate: Packages for helium_gateway found, but incompatible with the architectures configured
 * opkg_install_cmd: Cannot install package helium_gateway.
root@OpenWrt-G220:~# 

What's the next step for us to do? Thank you!

shawaj:
@JackeyTY can't test it as don't have the hardware but created a build for your target https://github.com/helium/gateway-rs/pull/84

JackeyTY:
Thanks! Actually I have one question, how to generate keys or how is it provisioned?

madninja:
> Thanks! Actually I have one question, how to generate keys or how is it provisioned?

The code does not yet support a secured key which is why they're dataonly for now. Once the code supports secured keys (ecc608) the keys will be generated/provisioned the same way as on current full hotspots

JackeyTY:
Got it, so should be using gateway-mfr-rs for provisioning keys on light hotspots. Thanks!

shawaj:
@JackeyTY you said your target was `arm-openwrt-linux-muslgnueabi` - what is the actual architecture of your hardware?

https://doc.rust-lang.org/nightly/rustc/platform-support.html

shawaj:
ping @JackeyTY - any update?

madninja:
I'm going to close this due to lack of updates

PR #84, opened 2021-08-12
target: add support for arm-openwrt-linux-muslgnueabi target

shawaj:
add build for arm-openwrt-linux-muslgnueabi target for bobcat light gateway

Relates-to: #80
Signed-off-by: Aaron Shaw <shawaj@gmail.com>


madninja:
@shawaj Isn't openwrt maybe a bit too generic? All the other platforms represent a hardware/os target whereas openwrt is just the OS target and is assuming arm/muslgnueabi

shawaj:
@madninja yeah probably. Maybe wait for @JackeyTY to test and then just change it to bobcat?

Or do you have another suggestion for the name?

madninja:
> @madninja yeah probably. Maybe wait for @JackeyTY to test and then just change it to bobcat?
> 
> Or do you have another suggestion for the name?

I would suggest changing it to bobcat so we can keep moving ahead? @JackeyTY can change it when they get a chance

JackeyTY:
so just changing the name to arm-bobcat-linux-muslgnueabi?

madninja:
> so just changing the name to arm-bobcat-linux-muslgnueabi?

no change the name to `bobcat` since we name platforms as the overarching name without the specific build triplet

madninja:
> @madninja yeah probably. Maybe wait for @JackeyTY to test and then just change it to bobcat?
> 
> Or do you have another suggestion for the name?

Would you be able to change the name to bobcat in your PR here @shawaj ?

shawaj:
Will do this evening 👍

shawaj:
@madninja all updated now

madninja:
@shawaj it looks like the target triplet is not correctly specced here. Did you use a correct cross variant?

madninja:
@shawaj are we still interested in this PR? May as well close I if it is no longer needed

shawaj:
Will close, easy enough to reopen similar if necessary

PR #104, opened 2021-09-21
Update settings.rs

topmop2:
loopback IP address modified


madninja (approved):
I can’t believe that slipped in. Thank you

PR #106, opened 2021-09-21
target: add support for mipsel_24kc architecture with mipsel-unknown-linux-musl target

greenbreakfast:
Adding support + build for mipsel_24kc architecture with mipsel-unknown-linux-musl target


madninja:
We prefer the targets to be actual product or product family names that are being targeted instead of cpu targets. This is because a target is an actual fully installable release which is more than just a cpu

greenbreakfast:
@madninja Could you clarify if you're referring to the `mipsel_24kc` architecture or the `mipsel-unknown-linux-musl` target?

If it's the architecture, would it suffice to change the architecture name to `smartharvest`? We chose `mipsel_24kc` based on the package architecture of our Linux module but can be flexible.

madninja:
> @madninja Could you clarify if you're referring to the `mipsel_24kc` architecture or the `mipsel-unknown-linux-musl` target?
> 
> If it's the architecture, would it suffice to change the architecture name to `smartharvest`? We chose `mipsel_24kc` based on the package architecture of our Linux module but can be flexible.

I'm referring to the target name which is `mipsel_24kc`. If you could rename that closer to the actual product name like smartharvest so that people can actually find it when they're looking for it?

greenbreakfast:
@madninja That makes sense, thanks for clarifying.

Updated the PR to set `smartharvest` as the target name!

madninja:
Excellent, thank you! PR a better checkmark once you confirm the build works?

madninja (approved):

greenbreakfast:
@madninja thanks! 

> PR a better checkmark once you confirm the build works?

Our local build works 100%. Should I wait for an official build + release before submitting a PR for the checkmark?

madninja:
nope.. I trust you'll actually test a build when it happens

PR #109, opened 2021-10-05
Implement state channels

madninja:
This switches the gateway to use state channels to offer packet hashes and only send the packet on successful purchase. In addition the gateway will attempt to detect conflicts and react to various close messages from the validator (called gateway service) to send either normal or disputed closes when the target router has not done so


lthiery:
Does this maintain free delivery of packets if the gateway/hotspot is not on-chain (cannot be paid)?

PR #111, opened 2021-10-08
create a Docker container

lthiery:


Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# Cargo Build Stage
+# ------------------------------------------------------------------------------
+
+FROM rust:latest as cargo-build
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y cmake musl-tools
+
+RUN rustup default nightly
```
madninja, line 11: Does this actually need nightly? I guess so for the strip features?

lthiery: Yes, I think it's for the strip features that I needed it. Similar to the other builds?

https://github.com/helium/gateway-rs/blob/main/.github/workflows/ci.yml#L18

madninja (approved):

PR #127, opened 2021-10-21
Add keypair ecc instructions

madninja:


madninja:
how's this look @shawaj ?

README.md
```diff
@@ -222,6 +222,20 @@ store = "/etc/helium_gateway/cache"
...
+
+If your gateway is enabled with an ECC608 crypto chip which is set up correctly, you can configure helium_gateway to use the crypto chip for secure key storage and crypto operations.  
+
+To use In your `settings.toml` override the `keypair` seting to reflect the use of the ECC and specify the bus address and slot to use. For example: 
+
+```
+keypair = "ecc://i2c-1:96&slot=0"
+```
+
+will have helium_gateway use the ECC at the `/dev/i2c-1` device driver location, use bus address `96` and slot `0` for it's crypto operations. Note that the file based keypair will no longer be used once the ECC is configured for use.
```
shawaj, line 235: the only thing that i think might be worth mentioning is that this is a hex address so corresponds to 0x60 on i2cdetect

PR #133, opened 2021-10-27
package: add mipsel_24kc, 1:1 of ramip_24kec

kallisti5:


madninja:
e.g. smartharvest updated their specific build to include the ecc specifier since their hardware has an ecc setup on board. I guess this common "mipsel" package stays common until it's not then the vendor has to create a new patch for a specific build/version?

kallisti5:
@madninja oh... did you ever run this on a rak device?

```
# opkg print-architecture
arch all 1
arch noarch 1
arch mipsel_24kc 10
```

If rak has ```mipsel_24kc``` and ```ramip_24kec``` in the output of the command above, we could consolidate these down into just ```mipsel_24kc``` and have it function universally.  The ```opkg print-architecture``` is what's key.

Sorry if I already asked this, been a while on this one.


I'll update the readme

madninja:
> @madninja oh... did you ever run this on a rak device?
> 
> ```
> # opkg print-architecture
> arch all 1
> arch noarch 1
> arch mipsel_24kc 10
> ```
> 
> If rak has `mipsel_24kc` and `ramip_24kec` in the output of the command above, we could consolidate these down into just `mipsel_24kc` and have it function universally. The `opkg print-architecture` is what's key.
> 

Yes.. see the [supported platforms](https://github.com/helium/gateway-rs#supported-platforms). We tested gateway-rs on some pretty old RAK EVB833 kits that still used the ramips_24kec architecture. I'd love to consolidate, just not sure that's possible for that particular piece of hardware and the OS it runs

fxntester:
My hardware platform is MIPS (Mediatek/MT7688). I have the same issue after I upgraded it
to latest OpenWRT (from V15 to V21). I observed the “mipsel_24kc” shall be main stream and 
right candidate. The ramip_24kec is quite old actually.

Would you decide to add "mipsel_24kc" in supported platform? Thanks.

root@Hotspot:/# uname -a
Linux Hotspot 5.4.154 #0 Sun Oct 24 09:01:35 2021 mips GNU/Linux
root@Hotspot:/# opkg print-architecture
arch all 1
arch noarch 1
arch mipsel_24kc 10

PR #135, opened 2021-10-28
smartharvest target: settings change to support ECC chip

greenbreakfast:
Updating Smart Harvest target settings (`package/smartharvest/data/etc/helium_gateway/settings.toml`) to override the keypair setting to use the ECC chip


greenbreakfast:
thanks @madninja

PR #149, opened 2021-11-11
add mipsel_24kc

Wvirgil123:
For ramips, the new openwrt uses "mipsel_24kc" as the Package architecture


madninja:
1. Is this a duplicate PR of #133 
2. If RAK (ramips) is now using mipsel_24kc why did you not remove ` ramips_24kec` from the build matrix? Can you confirm/show that ramips_24kec is no longer needed?

Wvirgil123:
> 1. Is this a duplicate PR of [package: add mipsel_24kc, 1:1 of ramip_24kec #133](https://github.com/helium/gateway-rs/pull/133)
> 2. If RAK (ramips) is now using mipsel_24kc why did you not remove ` ramips_24kec` from the build matrix? Can you confirm/show that ramips_24kec is no longer needed?

1. yes，I didn't notice this PR before
2. The new openwrt (such as openwrt 21) package architecture using ` mipsel_24kc`, but the old openwrt uses ` ramips_24kec`. For ramips to be compatible with all versions of openwrt, both of these are reserved.

madninja:
I'll close this in favor of #133

PR #150, opened 2021-11-11
Easier starting docker specified by Dockerfile

djieno:
To be able to start gateway-rs in a docker, a docker-compose is a welcome addition. I will also create the action to build the docker and host at dockerhub.

image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,6 +8,7 @@ services:
     environment:
       - GW_LOG_LEVEL=info
       - GW_LOG_METHOD=stdio
-      - GW_REGION=EU868
-      - GW_TZ=Europe/Amsterdam
+      - GW_LOG_TIMESTAMP=false # Change this to true in a debug/tcpdump situation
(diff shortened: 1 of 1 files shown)


madninja:
Thanks for starting this. Couple of comments

1. We host our images on quay, and will get the gateway-rs CI to build docker images there
2. While I like the docker compose option, this PR makes it look like it's only going to work if you're in the EU, which is not quite going to work as a general solution?

djieno:
@madninja  

1)Yes, that would be great. The image address could just be quai.io/gateway-rs:[tag]
2)I've created an update https://github.com/helium/gateway-rs/pull/150/commits/d7cada1e25c66a522de029e3789fb9e7e9d804af which might be more self explanatory. Instead of changing the settings.toml simply over rule the env variable

madninja:
> @madninja
> 
> 1)Yes, that would be great. The image address could just be quai.io/gateway-rs:[tag] 2)I've created an update [d7cada1](https://github.com/helium/gateway-rs/commit/d7cada1e25c66a522de029e3789fb9e7e9d804af) which might be more self explanatory. Instead of changing the settings.toml simply overrule using the env variable

Been a while, sorry. Not sure if this is still useful to you. If so you'll want to turn of updates in the compose as well. No need to do self upgrade checks since they'll be coming through docker presumably

madninja:
Closing due to lack of progress

PR #155, opened 2021-11-30
create mod subnet

mikev:


lorawan/src/subnet.rs
```diff
@@ -0,0 +1,386 @@
...
+    let netid = parse_netid(devaddr);
+    let (lower, _upper) = netid_addr_range(netid, netid_list);
+    lower + nwk_addr(devaddr)
+}
+
+fn netid_class(netid: u32) -> u32 {
+    netid >> 21
+}
+
+fn addr_len(netclass: u32) -> u32 {
```
JayKickliter, line 22:
```rust
fn addr_len(netclass: u32) -> u32 {
    *[25, 24, 20, 17, 15, 13, 10, 7]
        .get(netclass as usize)
        .unwrap_or(&0)
}
```

lorawan/src/subnet.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+    0
+}
+
+fn subnet_addr_within_range(subnetaddr: u32, netid: u32, netid_list: Vec<u32>) -> bool {
+    let (lower, upper) = netid_addr_range(netid, netid_list);
+    (subnetaddr >= lower) && (subnetaddr < upper)
+}
+
+fn var_net_class(netclass: u32) -> u32 {
```
lthiery, line 72: would it be worth making a special enum type for netclass?
```rust
enum NetClass {
  _0,
  ...
  _7,
}
```

lthiery (commented):
I would recommend separating out LoRaWAN from the Helium specific routing topics.

For example, all the functions that takes NetID List seem pretty domain specific and probably shouldn't be here (perhaps in the router module somehwere).

However, many of these other helpers do seem relevant in a lorawan module.

lthiery (commented):
Would a lot of this be nicer with a DevAddr type?

Then a lot of these functions could hang off of the implementation.

JayKickliter:
> Would a lot of this be nicer with a DevAddr type?

I was wondering the same thing, but it feels a little pedantic. I guess it depends on @mikev background with rust. If new, then too much typing is overload

lthiery:
> If new, then too much typing is overload

Totally fair. I just love to type all the things at this point with Rust but in the beginning it doesn't feel as compelling.

lorawan/src/subnet.rs
mikev: Hi Jay - This is my first attempt at Rust code.  You'll have to explain this line by line :-)

lorawan/src/subnet.rs
mikev: Technically the netclass is a 3 bit value, i.e. u3.  I explored the enum idea, but netclass is used in different ways in the code, i.e. a number, a bitshift input value and bit pattern.  The code starts looking ugly when it is declared an enum.

I changed it to a u8.  The usage is safer now.  What do you think?

mikev: > For example, all the functions that takes NetID List seem pretty domain specific and probably shouldn't be here (perhaps in the router module somehwere).

Good point and I agree with the domain specific premise.  Pragmatically though this same logic is shared between the light gateway, Miner, Router and PacketPurchaser repos.  Prior reviewers noted the duplicate code and consensus was to put the subnet translation logic into the shared LoRaWAN library.

lorawan/src/subnet.rs
```diff
@@ -0,0 +1,389 @@
...
+        1 => 0b10u32 << idlen,
+        2 => 0b110u32 << idlen,
+        3 => 0b1110u32 << idlen,
+        4 => 0b11110u32 << idlen,
+        5 => 0b111110u32 << idlen,
+        6 => 0b1111110u32 << idlen,
+        7 => 0b11111110u32 << idlen,
+        _ => 0,
+    };
+    result
```
JayKickliter, line 86: FYI: everything in rust is an expression, so you can write this as only the match:
```
fn var_net_class(netclass: u8) -> u32 {
    let idlen = id_len(netclass);
    match netclass {
         0 => 0,
         1 => 0b10u32 << idlen,
         2 => 0b110u32 << idlen,
         3 => 0b1110u32 << idlen,
         4 => 0b11110u32 << idlen,
         5 => 0b111110u32 << idlen,
         6 => 0b1111110u32 << idlen,
         7 => 0b11111110u32 << idlen,
         _ => 0,
    } // <- notice lack of semicolon. removing it lets the the expression (match block) value fall out
}
```

lorawan/src/subnet.rs
JayKickliter: No prob. We can pair if it would help

lorawan/src/subnet.rs
mikev: F# has an amazing type system, where for example, a custom type can be defined for two different unsigned ints and the compiler will enforce it.  The meter to feet conversion is the common example where you want to declare a distance in meters and distance in feet types and let the compiler strictly enforce it, so mistaken assignments do not occur.

Prof pointed out that something similar may exist for Rust
https://yoric.github.io/post/uom.rs/

madninja: yeah.. I wouldn't try to overcomplicate things though

lorawan/src/subnet.rs
mikev: I read up on the Rust raw pointer.  I think I understand now; this is useful to know.  Seems reasonable; updated.

dpezely (approved):
Be sure to run [Clippy](https://github.com/rust-lang/rust-clippy), but addressing its recommendations can be a future PR.  I continue learning much from what Clippy suggests.

JayKickliter:
> Be sure to run [Clippy](https://github.com/rust-lang/rust-clippy), but addressing its recommendations can be a future PR.

We generally consider any clippy warnings a blocker for merging. Not sure if this repos CI enforces it

madninja:
@mikev please remove println from code. It is not good practice to leave those sprinkled around since they don’t always go where they should

PR #162, opened 2021-12-20
Add LongAP Pro using ATECC608

timcooijmans:
There will be multiple security chips used by LongAP Pro, so we have to be specific per security chip


madninja:
> There will be multiple security chips used by LongAP Pro, so we have to be specific per security chip

I'd propose you add specific model numbers/names for each hardware variant/skew?

README.md
```diff
@@ -81,23 +81,24 @@ The following platforms have already been tested by Helium and our community. Ou
...
+| dragino          | mips-unknown-linux-musl        | :white_check_mark: Dragino [LPS8]                                                                                           |
+|                  |                                | :grey_question: Dragino [DLOS8]                                                                                             |
+| mtcdt            | armv5te-unknown-linux-musleabi | :white_check_mark: Multitech Conduit [MTCDT] (mLinux)                                                                       |
+| resiot           | armv7-unknown-linux-gnueabihf  | :white_check_mark: ResIOT X gateways [resiot]                                                                               |
+| cotx             | aarch64-unknown-linux-gnu      | :white_check_mark: Cotx gateways [cotx]                                                                                     |
+| longap_pro_atecc | mipsel-unknown-linux-musl      | :white_check_mark: LongAP Pro (using ATECC608)                                                                              |
```
madninja, line 97: I suggest adding the specific model name/version to this README.. people won't know which is has the ECC version otherwise?

timcooijmans: > I suggest adding the specific model name/version to this README.. people won't know which is has the ECC version otherwise?

The right version (compiled by us) will be installed by our software management/update system, we don’t expect customers to download and install gateway_rs themselves. This PR is more because we are required to by DeWi Audit requirements.

madninja: Gotcha.. In that case it doesn't have to be in the README either. Nice for some flair I guess. 

If you're managing your own updates too instead of relying on the self-updater, also turn of the updater btw (most docker implementations would likely do this). It's fine to reflect that in this PR if you already do

timcooijmans: It has to be in the README according to the DeWi audit requirements. I could have just added our hotspots to the README list without a build, but that feels a like cheating. But yes this is kind of a throw away build.

timcooijmans: Do you want me to disable updating in the installed default config because that reflects our actual install a bit better?

madninja: yes please, _and_ actually reflect what your build does? Because this build will not work on your hardware as is without an i2c key setting right?

madninja: hmm.. so this build is technically not for your hardware, or your ecc.. it's literally a standard non ecc mipsel build. I don't see how this helps DeWi at all.

timcooijmans: Oh, I got the issue, the config file I intended to add was ignored by git ignore. I sec

timcooijmans: Added the config-file. So to clarify, this build works on our hardware, uses the security IC, but are building and releasing it ourselves.

README.md
```diff
@@ -94,7 +94,7 @@ Note that platforms will be tested much faster if you join the development proce
 | mtcdt            | armv5te-unknown-linux-musleabi | :white_check_mark: Multitech Conduit [MTCDT] (mLinux)                                                                       |
 | resiot           | armv7-unknown-linux-gnueabihf  | :white_check_mark: ResIOT X gateways [resiot]                                                                               |
 | cotx             | aarch64-unknown-linux-gnu      | :white_check_mark: Cotx gateways [cotx]                                                                                     |
-| longap_pro_atecc | mipsel-unknown-linux-musl      | :white_check_mark: LongAP Pro (using ATECC608)                                                                              |
+| longap_pro_atecc | mipsel-unknown-linux-musl      | :white_check_mark: [LongAP Pro] (using ATECC608 security IC)                                                                |
```
madninja, line 97: no need to add the  "security IC"

timcooijmans: Gone!

timcooijmans:
Closing PR as adding hotspots to the README just for DeWi auditing purposes is no longer required.

PR #175, opened 2022-01-18
convert configurable endpoint to port only

jeffgrunewald:
Making the api endpoint the gateway listens on completely configurable raises security concerns by potentially exposing the endpoint on a different listening address then the localhost. By restricting the configurability to the listening port only this allows for running the gateway alongside other services that default to the same (4467) port as well as multiple gateways on the same host (typically for testing purposes).

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1,8 +1,7 @@
 mod client;
 mod server;
 
-const LISTEN_ADDR: &str = "127.0.0.1:";
-const URI_PREFIX: &str = "http://";
+const LISTEN_ADDR: &str = "127.0.0.1";
 
 pub use client::LocalClient;
 pub use helium_proto::services::local::{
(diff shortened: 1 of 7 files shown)


src/api/mod.rs
```diff
@@ -1,9 +1,25 @@
...
+
 pub use client::LocalClient;
 pub use helium_proto::services::local::{
     ConfigReq, ConfigRes, ConfigValue, EcdhReq, EcdhRes, HeightReq, HeightRes, PubkeyReq,
     PubkeyRes, SignReq, SignRes,
 };
 pub use server::LocalServer;
+
+pub fn listen_addr(port: u16) -> String {
+    let mut address = LISTEN_ADDR.to_string();
```
madninja, line 15: use the format! macro

src/api/mod.rs
```diff
@@ -1,9 +1,25 @@
...
+    address += &port.to_string();
+    address
+}
+
+pub fn connect_uri(port: u16) -> String {
+    let mut uri = URI_PREFIX.to_string();
```
madninja, line 21: use the format! macro

src/api/mod.rs
madninja: like `format!("{}:{}", LISTEN_ADDR, port)` and define LISTEN_ADDR as just 127.0.0.1

src/api/mod.rs
madninja: like `format!("http://{}", listen_addr(port))`

madninja (approved):

PR #184, opened 2022-02-01
Fix some typos

lthiery:


madninja (commented):
Looks like clippy caught up to something. Can you fix that clippy error too?

madninja (approved):

PR #186, opened 2022-02-08
Platform Addition - Tektelic Gateways

PeterHeckel-TEK:
The goal of this PR is to add Tektelic to the list of Helium supported gateways. Files to build a Tektelic ipk have been added and the Makefile has been updated to support that. 

helium-crypto-rs has also been upgraded to include the support of the SWI ECC chip, necessary for operation.


Makefile.toml
```diff
@@ -113,7 +118,7 @@ script = '''
         envsubst < config/conffiles > ${PKG_BUILD_DIR}/control/conffiles
         chmod +x ${PKG_BUILD_DIR}/control/conffiles
 	for control_file in control preinst postinst prerm; do
-	    envsubst < ${PKG_SRC_DIR}/control/${control_file} > ${PKG_BUILD_DIR}/control/${control_file}
+	    envsubst "$(printf '${%s} ' $(env | sed 's/=.*//'))" < ${PKG_SRC_DIR}/control/${control_file} > ${PKG_BUILD_DIR}/control/${control_file}
```
PeterHeckel-TEK, line 121: This change was made to support the use of bash variables in the control scripts. Without this change, `envsubst` will remove all local bash variables as it is trying to substitute  these local variables with non-existent environment variables. With this change, the local variables are retained if they are not substituted by `envsubst`

PeterHeckel-TEK:
PR not required to pass HW audit due to tektelic managing releases & upgrading

madninja:
> PR not required to pass HW audit due to tektelic managing releases & upgrading

👍🏻 make sure to disable update checking in your default/settings.toml to stop the update checker from running

package/tektelic/data/etc/helium_gateway/settings.toml
```diff
@@ -0,0 +1,90 @@
...
+## can support it.
+##
+## Override the setting by putting a keypair line in the 
+## local settings.toml. Changing things in this file will be overwritten
+## by upgrades. 
+##
+## File: 
+#keypair = "/etc/helium_gateway/gateway_key.bin"
+## ECC608 based
+keypair = "ecc://ttyO1:0&slot=0"
```
madninja, line 18: Fix this to be `ecc://ttyO1:0?slot=0`

package/tektelic/data/etc/helium_gateway/settings.toml
```diff
@@ -0,0 +1,90 @@
...
+method = "syslog"
+level = "info"
+timestamp = false
+
+[update]
+# Disable update checking
+enabled = false
+# OVERRIDE: Platform to use as part of update check. This MUST be overriden by a
+# local settings file
+platform = "unknown"
```
madninja, line 32: Please remove from here on downwards.. default.toml already includes all these settings so there is no need to override them

PeterHeckel-TEK:
All comments have been resolved

PR #192, opened 2022-02-14
Use default macaddr

madninja:


lthiery (approved):
That is much more pleasant.

PR #193, opened 2022-02-15
Adjust region based on blockchain

madninja:
This sets the gateway to follow the asserted lorawan region by following the regions through the associated gateway-service. 

The helium-gateway cli adds a new `region` info argument which will fetch the currently set region for the gateway. 

**Note**

The region setting in `settings.toml` will be used as the initial setting until a region update is received from the chain. This means that if the hotspot is never location asserted on chain the region will remain set to the settings.toml value. 

The info region command will return the default setting if no chain setting is asserted or found

### DIFF
--- a/src/router/dispatcher.rs
+++ b/src/router/dispatcher.rs
@@ -133,7 +133,8 @@ impl Dispatcher {
 
     pub async fn run(&mut self, shutdown: triggered::Listener, logger: &Logger) -> Result {
         let logger = logger.new(o!("module" => "dispatcher"));
-        info!(logger, "starting");
+        info!(logger, "starting"; 
+            "region" => self.region.to_string());
 
         info!(logger, "default router";
             "pubkey" => self.default_router.pubkey.to_string(),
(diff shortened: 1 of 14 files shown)


src/router/dispatcher.rs
```diff
@@ -242,6 +278,31 @@ impl Dispatcher {
...
+        &mut self,
+        response: &service::gateway::Response,
+        logger: &Logger,
+    ) {
+        let update_height = response.height();
+        if update_height <= self.region_height {
+            warn!(
+                logger,
+                "region returned invalid height {} while at {}", update_height, self.region_height
+            );
```
jeffgrunewald, line 291:
```suggestion
        let region_height = self.region_height;
        if update_height <= region_height {
            warn!(logger, "region returned invalid height {update_height} while at {region_height}");
```

src/router/dispatcher.rs
```diff
@@ -273,14 +331,11 @@ impl Dispatcher {
...
-                Err(err) => warn!(logger, "failed to parse routing: {:?}", err),
+                Err(err) => warn!(logger, "failed to parse routing: {err:?}"),
             }
         }
         self.routing_height = update_height;
-        info!(
-            logger,
-            "updated routing to height {:?}", self.routing_height
-        )
+        info!(logger, "updated routing to height {:?}", update_height)
```
jeffgrunewald, line 338:
```suggestion
        info!(logger, "updated routing to height {update_height:?}")
```

Issue #203, opened 2022-02-28
Issue Pulling The Status On My Dragino LPS8 915


dcgtechnologies:
Hello....I am trying to get my DATA Only gateway registered on the network and after I run the command "helium_gateway key info". I get the error below:

Error: Service(Service(tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 146, kind: ConnectionRefused, message: "Connection refused" })))))

I am including my logs below:

Mon Feb 28 10:22:00 2022 user.notice iot_keep_alive: Internet Access OK: via eth1
Mon Feb 28 10:22:00 2022 user.notice iot_keep_alive: use WAN or WiFi for internet access now
Mon Feb 28 10:22:15 2022 user.notice iot_keep_alive: Internet Access OK: via eth1
Mon Feb 28 10:22:15 2022 user.notice iot_keep_alive: use WAN or WiFi for internet access now
Mon Feb 28 10:22:20 2022 daemon.err uhttpd[4050]: wc: /tmp/logfile.log: No such file or directory
Mon Feb 28 10:22:20 2022 daemon.err uhttpd[4050]: sh: 2000000: unknown operand
Mon Feb 28 10:22:23 2022 daemon.err uhttpd[4050]: wc: /tmp/logfile.log: No such file or directory
Mon Feb 28 10:22:23 2022 daemon.err uhttpd[4050]: sh: 2000000: unknown operand
Mon Feb 28 10:22:23 2022 daemon.err uhttpd[4050]: killall: logread: no process killed
Mon Feb 28 10:22:26 2022 daemon.info fwd[20924]:
Mon Feb 28 10:22:26 2022 daemon.info fwd[20924]: ################[PKT_SERV] no report of this service ###############
Mon Feb 28 10:22:26 2022 daemon.info fwd[20924]: PKTUP~ [server2] JSON: {"stat":{"time":"2022-02-28 16:22:26 UTC","rxnb":0,"rxok":0,"rxfw":0,"ackr":0.0,"dwnb":0,"txnb":0,"pfrm":"SX1308","mail":"dragino-1f7e68@dragino.com","desc":"Dragino LoRaWAN Gateway"}}
[...]

Issue #202, opened 2022-02-28
 1.0.0-alpha.22 info error


dvdadie2:
when this command is issued:  helium_gateway info

this is returned:

Error: Service(Service(tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 146, kind: ConnectionRefused, message: "Connection refused" })))))

does anyone know how to fix this from the ssh command line?
i found some commands in the code - "start_service", "enable", "disable","reload" but those don't appear to be useful when used after "helium_gateway".

lthiery:
You need to run `helium_gateway start_service` so that the service begins. When the service is running, 
 you can run `helium_gateway info`.

dvdadie2:
@lthiery thank you.  i really wish that would have worked, but it didn't.  

instead of starting the service, this is returned:
error: Found argument 'start_service' which wasn't expected, or isn't valid in this context

some background - i am using the dragino lps8 and recently upgraded to firmware version lgw-5.4.1644990565.  i believe that during that upgrade the helium_gateway version updated from alpha.21 (which worked) to alpha.22 (not currently working).  my suspicion is that something is wrong with the dragino file(s) and / or the dragino GUI.

please note that i am not looking for dragino help - i am just looking for gateway-rs help.  
that is why i am trying to bypass the dragino GUI and control the unit from putty / ssh.

i tried to downgrade the helium_gateway version from alpha.22 to alpha.21 but it wouldn't let me.
i also tried to re-install alpha.22 with this command:

opkg install /tmp/helium-gateway-v1.0.0-alpha.22-dragino.ipk

which returned:

Package helium_gateway (1.0.0-alpha.22) installed in root is up to date.

nothing noticeably changed.

thanks for your time.

drberthol:
same with lig16 since few hours....

lthiery:
I was wrong about "start_service". It's just `helium_gateway server`. Also, you can access the help menu by using `--help` to see valid commands and explanations of how specific commands work.

eg:
```
helium_gateway --help
helium_gateway server --help
```

lthiery:
You can ignore those JSON parsing errors but I'll reach out to Dragino about that anyway.

[...]

PR #204, opened 2022-02-28
Improve Error Msgs

lthiery:
#203 #202 are similar tickets where the user was not running `helium_gateway server` before trying to use the info command.

The cited error message is a little cryptic for a non-dev:
```
Error: Service(Service(tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 146, kind: ConnectionRefused, message: "Connection refused" })))))
```

The changes in `main.rs` use the logger to output the error before shutting down, thus using the custom display in the error's attribute:
```
Feb 28 19:14:36.222 ERRO service error service transport error
```

Adding colons and debug to the lowest level error message yields somethiong similar to the original error message,  but a little nicer:

```
Feb 28 19:18:47.945 ERRO service error: service: tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })))
```

Mapping the transport error during `LocalClient::new` provides an explicit hint to the user:
```
Feb 28 19:15:30.602 ERRO service error: Unable to connect to local client. Check the `helium_gateway server` is running.
```

### DIFF
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -1,7 +1,7 @@
 use super::{
     connect_uri, ConfigReq, ConfigValue, HeightReq, HeightRes, PubkeyReq, RegionReq, SignReq,
 };
-use crate::{error::ServiceError, PublicKey, Region, Result};
+use crate::{error::Error, PublicKey, Region, Result};
 use helium_proto::services::local::Client;
 use std::convert::TryFrom;
 use tonic::transport::{Channel, Endpoint};
@@ -16,7 +16,7 @@ impl LocalClient {
         let endpoint = Endpoint::from_shared(uri).unwrap();
         let client = Client::connect(endpoint)
             .await
-            .map_err(ServiceError::LocalClientConnect)?;
+            .map_err(Error::local_client_connect)?;
         Ok(Self { client })
     }
 
--- a/src/api/client.rs
+++ b/src/api/client.rs
@@ -1,7 +1,7 @@
 use super::{
     connect_uri, ConfigReq, ConfigValue, HeightReq, HeightRes, PubkeyReq, RegionReq, SignReq,
 };
-use crate::{PublicKey, Region, Result};
+use crate::{error::ServiceError, PublicKey, Region, Result};
 use helium_proto::services::local::Client;
 use std::convert::TryFrom;
 use tonic::transport::{Channel, Endpoint};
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -70,6 +70,8 @@ pub enum ServiceError {
     NoService,
     #[error("block age {block_age}s > {max_age}s")]
     Check { block_age: u64, max_age: u64 },
+    #[error("Unable to connect to local client. Check the `helium_gateway server` is running.")]
```
madninja, line 73:
```suggestion
    #[error("Unable to connect to local server. Check that `helium_gateway` is running.")]
```

src/error.rs
```diff
@@ -58,9 +58,9 @@ pub enum DecodeError {
 
 #[derive(Error, Debug)]
 pub enum ServiceError {
-    #[error("service {0}")]
+    #[error("service: {0:?}")]
```
madninja, line 61: why the gratuitous debug for errors? I actually have cleaner messages for some of these?

src/error.rs
```diff
@@ -18,7 +18,7 @@ pub enum Error {
     Encode(#[from] EncodeError),
     #[error("decode error")]
     Decode(#[from] DecodeError),
-    #[error("service error {0}")]
+    #[error("service error: {0}")]
```
madninja, line 21: but but... I had these lined up to like form sentences :-/

src/main.rs
```diff
@@ -107,8 +107,12 @@ pub fn main() -> Result {
     });
     runtime.shutdown_timeout(Duration::from_secs(0));
 
+    if let Err(e) = &res {
+        let error_logger = slog_scope::logger().new(o!());
+        error!(&error_logger, "{e}");
```
madninja, line 112: I think you can just use `logger` above? No need to create yet another scope.

src/api/client.rs
```diff
@@ -14,7 +14,9 @@ impl LocalClient {
     pub async fn new(port: u16) -> Result<Self> {
         let uri = connect_uri(port);
         let endpoint = Endpoint::from_shared(uri).unwrap();
-        let client = Client::connect(endpoint).await?;
+        let client = Client::connect(endpoint)
+            .await
+            .map_err(ServiceError::LocalClientConnect)?;
```
madninja, line 19: Use the pattern used in the crate.. methods to construct errors, please

src/main.rs
madninja: I believe you can just use `logger` or worst case clone it.. no need to re-scope one

PR #218, opened 2022-03-10
target: adding support for ClodPi light hotspot

karthiclodpi:
Adding support for ClodPi light hotspot


madninja:
1. YOu're missing settings.toml updates for ECC support? 
2. Are you actually going to use the built-in update mechanism? If not there is no need for this PR since you can pass HIP19 approval with just gateway-mfr-rs.

karthiclodpi:
Hi @madninja 

we have tested with ClodPi Light hotspot. we want  to use the built-in update mechanism.

added settings.toml

PR #219, opened 2022-03-12
Madninja/use slice

madninja:


mikev (approved):

lorawan/src/subnet.rs
```diff
@@ -47,16 +47,14 @@ fn id_len(netclass: u8) -> u32 {
...
-        if subnet_addr_within_range(subnetaddr, item, netid_list.clone()) {
-            return item;
-        }
-    }
-    0
+fn subnet_addr_to_netid(subnetaddr: u32, netid_list: &[u32]) -> u32 {
+    *netid_list
+        .iter()
+        .find(|item| subnet_addr_within_range(subnetaddr, **item, netid_list))
+        .unwrap_or(&0)
```
JayKickliter, line 54: Is 0 the correct default case here, or should it be None?

madninja: 0 is what the original code did. I’d return an Option but that may not match the code in the spec? @mikev ?

mikev: Rust has an Option?  Nice, I did not know that.  I'll revisit the return type for this function.

madninja: Sure does. I'd also actually make types for netid/devaddr instead of just returning u32 for both to make it clearer what you're talking about. You can then also tie functions to those types which may make this all clearer to understand.

lorawan/src/subnet.rs
```diff
@@ -86,7 +86,12 @@ fn devaddr(netid: u32, nwkaddr: u32) -> u32 {
 }
 
 fn is_local_netid(netid: u32, netid_list: &[u32]) -> bool {
-    netid_list.contains(&netid)
+    let retired_netid: u32 = 0x200010;
```
madninja, line 89: make this a const

mikev: Done

mikev:
FYI - I created this branch to address the types and Options comments:
https://github.com/helium/gateway-rs/tree/mv/subnet_v2

PR #225, opened 2022-03-22
bump semtech-udp

madninja:
This also pins the underlying log crate using `cargo update -p log --precise 0.4.14` to avoid `slot-stdlog` from picking up a breaking private api change. 

This update also indirectly updates `ecc-608-linux-rs` since it's a floating dependency. Pulling in the latest version `v0.1.7` should fix some interaction issues a few i2c implementations have seen


madninja:
Superseded by #226

PR #227, opened 2022-03-24
example docker-compose.yml file

joecryptotoo:
\run ```docker-compose build``` and then ```docker-compose up```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,8 +8,8 @@ services:
     image: helium/gateway-rs
     build:
       context: .
-    #volumes:
-    #  - './gateway_key.bin:/etc/helium_gateway/gateway_key.bin:ro'
+    volumes:
+      - './gateway_key.bin:/etc/helium_gateway/gateway_key.bin'
     #  - '/dev/i2c-1:/dev/i2c-1'
     #cap_add:
     #  - SYS_RAWIO
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,8 +8,11 @@ services:
     image: helium/gateway-rs
     build:
       context: .
-    volumes:
-      - './gateway_key.bin:/etc/helium_gateway/gateway_key.bin:ro'
+    #volumes:
+    #  - './gateway_key.bin:/etc/helium_gateway/gateway_key.bin:ro'
+    #  - '/dev/i2c-1:/dev/i2c-1'
+    #cap_add:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+      - GW_LOG_LEVEL=info 
+      - GW_UPDATE_ENABLED=false
+    networks:
+      - helium
+    expose:
+      - 1680/udp
+      - 4467/tcp
+    ports:
+      - 1680:1680/udp
+      - 4467:4467/tcp
```
madninja, line 24: I'm not sure what the right setup here is.. you _definitely_ do _not_ want 4467 (the local api) exposed to the open internet since it basically allows anyone to get your gateway to sign anything.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3.7"
+
+networks:
+  helium:
+
+services:
+  gateway-rs:
+    image: helium/gateway-rs:v1.0.0-alpha.23
+    build:
+      context: https://github.com/helium/gateway-rs.git#v1.0.0-alpha.23
```
madninja, line 10: How are we going to keep this up to date?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: "3.7"
+
+networks:
+  helium:
+
+services:
+  gateway-rs:
+    image: helium/gateway-rs:v1.0.0-alpha.23
```
madninja, line 8: Who is going to keep updating this?

madninja (commented):
Well done, but it seems like this would have some maintenance overhead and is tailored to your preference? Should this be maintained in this repo at all?

docker-compose.yml
joecryptotoo: We can change the tag to latest and just have it build main

docker-compose.yml
joecryptotoo: I changed it to bind to localhost on the host machine.

joecryptotoo: https://github.com/helium/gateway-rs/pull/227/commits/5760d9c21cc41a2d8fa6456f64689df6f73c96de

docker-compose.yml
joecryptotoo: I changed it to just build the local source.

madninja:
Closing due to lack of response.

Issue #230, opened 2022-03-30
Gateway connection refused


Oliv4945:
Dear,

Since few hours my 3 different data only hotspots (connected on 3 distinct networks) are all reporting the error:
```
gateway error: Service(Rpc(Status { code: Unavailable, message: "error trying to connect: tcp connect error: Connection refused (os error 111)", source: Some(tonic::transport::Error(Transport, hyper::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })))) })), module: dispatcher
```
Are the gateway working correctly?

Additionally the gateways in `default.toml` file are not responding to ping, but it might be disabled?

Full log:
```
Mar 30 18:16:10 raspberrypi systemd[1]: helium_gateway.service: Main process exited, code=killed, status=15/TERM
Mar 30 18:16:10 raspberrypi systemd[1]: helium_gateway.service: Succeeded.
Mar 30 18:16:10 raspberrypi systemd[1]: Stopped Helium Gateway.
Mar 30 18:16:10 raspberrypi systemd[1]: Started Helium Gateway.
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: starting server, key: 13EvWRj72dJFUSNHZU1zi3oUVKdrgeNEWKDGzKCd6CYVtoduje5, version: 1.0.0-alpha.23
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: starting, module: gateway, listen: 127.0.0.1:1680
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: starting, module: dispatcher, region: EU868
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: default router, module: dispatcher, uri: http://54.176.88.149:8080/, pubkey: 1124CJ9yJaHq4D6ugyPCDnSBzQik61C1BqD9VMh1vsUmjwt16HNB
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: using gateway, module: dispatcher, uri: http://3.38.70.101:8080/, pubkey: 11yJXQPG9deHqvw2ac6VWtNP7gZj8X3t3Qb3Gqm9j729p4AsdaA
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: starting, module: updater
Mar 30 18:16:10 raspberrypi helium_gateway[1919]: starting, listen: 127.0.0.1:4467, module: api
[...]

PR #232, opened 2022-03-31
Diversify config with Argon validators

mbthiery:
In response to #230, this PR might help serve as a bridge until there's a system to include non-hardcoded validators. I've added a couple validators managed by Argon which adds diversification to the list, hopefully making an outage (even in alpha) less likely.

I host a few helium_gateway instances myself and I've tested these endpoints.


madninja:
Appreciated but see [this PR](https://github.com/helium/gateway-rs/pull/234) to use the static config as just a seed list to connect to actual staked validators

PR #240, opened 2022-04-05
Allow absent default routers

### LINKED ISSUE #239: Gateway shouldn't crash when router channel isn't defined
For router update channels like the newly added `testnet` channel, when the `router.testnet` uri and pubkey aren't defined in any settings config (default.toml or settings.toml, etc) the gateway should not immediately crash when it can't find the components to construct a KeyedUri for the default router on that chain.

The current behavior is that the process will immediately terminate with an error that `uri` is missing and will behave similarly if a uri is defined but a pubkey is not.

madninja:
Fixes #239

### DIFF
--- a/src/router/dispatcher.rs
+++ b/src/router/dispatcher.rs
@@ -88,10 +88,10 @@ pub struct Dispatcher {
     gateways: Vec<KeyedUri>,
     routing_height: u64,
     region_height: u64,
-    default_router: KeyedUri,
     cache_settings: CacheSettings,
-    routers: HashMap<RouterKey, RouterEntry>,
     gateway_retry: u32,
+    routers: HashMap<RouterKey, RouterEntry>,
+    default_router: Option<KeyedUri>,
 }
 
 #[derive(PartialEq, Eq, Hash)]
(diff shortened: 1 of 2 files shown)


src/router/dispatcher.rs
```diff
@@ -88,7 +88,7 @@ pub struct Dispatcher {
     gateways: Vec<KeyedUri>,
     routing_height: u64,
     region_height: u64,
-    default_router: KeyedUri,
+    default_router: Option<KeyedUri>,
```
JayKickliter, line 91: OCD: I prefer seeing Option fields at the bottom of a struct

madninja: OCD pacified!

src/settings.rs
```diff
@@ -111,8 +111,10 @@ impl Settings {
             .map_err(|e| e.into())
     }
 
-    pub fn default_router(&self) -> &KeyedUri {
-        &self.router[&self.update.channel.to_string()]
+    pub fn default_router(&self) -> Option<KeyedUri> {
+        self.router
+            .get(&self.update.channel.to_string())
+            .map(|uri| uri.to_owned())
```
jeffgrunewald, line 117: this seems like a misnomer since it's not just going to get the uri element of the router keyeduri

madninja: I've interchangeably used `uri` before to indicate `keyed_uri`. Can't disagree it's a bit ambiguous. We'll have to take a pass at clearing this up.

madninja: and fixed in 1da305351aa112d15c788078ccea559235500813

PR #250, opened 2022-05-13
define new Regions AS923_1B and CD900_1A

mikev:
AS923_1B defines a Region supporting Malaysia. It is illegal in Malaysia to use any frequency above 924 Mhz, hence the reason for this new Region. The AS923_1B defines Uplink frequencies at [922.0, 922.2, 922.4, 922.6, 922.8, 923.0, 923.2, 923.4]

CD900_1A defines a new testnet Region. This will be used for testing new Region parameters. It is centered in the ocean at the Challenger Deep which is the lowest sea floor depth on Earth - https://en.wikipedia.org/wiki/Challenger_Deep

CN779 is officially deprecated by the LoRa Alliance and should never be used.

PR is pending to update the proto Region enum
https://github.com/helium/proto/pull/132

Note, this change updates the Cargo.lock link to Helium.Proto.  I'll let madninga decide when this is OK to merge.  It does include several updates to Helium.Proto.


madninja (commented):
You can't just move the CN799 enum entry right? It's still going to be defined in

oberchoo:
Yes, Malaysia will need that in making Helium Network legal.

mikev:
I've added back CN779.  

I do believe we should sort out an update method to permanently remove some enum names.  Ideally we are going to rename some of these plan names to be more explanatory and less vague.  For example, US915 will become US915_SB2.  This will allow us to deploy any variant such as US915_SB1, US915_SB7, etc.  So, I'd like to see some method to be able to permanently remove names, otherwise we'll need to leave US915 in place forever which will be confusing.

madninja:
> I've added back CN779.
> 
> I do believe we should sort out an update method to permanently remove some enum names. Ideally we are going to rename some of these plan names to be more explanatory and less vague. For example, US915 will become US915_SB2. This will allow us to deploy any variant such as US915_SB1, US915_SB7, etc. So, I'd like to see some method to be able to permanently remove names, otherwise we'll need to leave US915 in place forever which will be confusing.


For backwards compatibility you can't really do that without a synchronized first upgrade where you break every one which is something we'd avoid. The _name_ can be changed however since that won't break the on the wire format, just the source builds. 

So we can rename US915 to US915_SB2 in the proto and it wouldn't break running systems with different older and newer versions of the proto

PR #259, opened 2022-06-01
Run update script as a fork to fix install issues

PaulVMo:
### Problem Statement
When running an update, the existing service never terminates the currently running instance of helium_gateway. By leaving the service running, it causes the gateway to be inoperable when the new version is installed. Rebooting the gateway resolves this. However, even with a reboot and the gateway functioning, the opkg package manager is out of sync. `opkg list` will continue to show the old version. This appears to be related to the fact that the update install script is running as part of the existing service.

### Solution
First, as part of the `prerm` (pre-removal) script that is called as part of an update, add a call to stop the current helium_gateway service. However, because this terminates the service that is running the upgrade, this causes it's own issues. So, secondly, this update also forks the process before calling the update install script. This ensures that the install script survives after the gateway service is stopped.

### Open Issues
Currently, the logging does not work in the forked child process. I am not sure if this is because the logger is async and the process terminates before allowing the logging to complete, or if the logger is just not accessible to the fork for some reason. I could use some help/insight into that. I tested this on a Dragino and RAK WisGate by rebasing this PR on the alpha.25 tag

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,7 @@ semtech-udp = { version = ">=0.9.4,<1", default-features=false, features=["serve
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.3.4", features = ["ecc608"]}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
+fork = "0.1"
 
 [profile.release]
 opt-level = "z"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ thiserror = "1.0"
 base64 = "0"
 rand = "0.8"
 prost = "0"
-daemonize = "0.4"
+daemonize = {git = "https://github.com/knsd/daemonize", rev = "f7be28e"}
 tonic = "0"
 http = "*"
 log = "0"
@@ -55,7 +55,6 @@ semtech-udp = { version = ">=0.9.4,<1", default-features=false, features=["serve
 helium-proto = { git = "https://github.com/helium/proto", branch="master", features=["services"]}
 helium-crypto = { git = "https://github.com/helium/helium-crypto-rs", tag = "v0.3.4", features = ["ecc608"]}
 longfi = { git = "https://github.com/helium/longfi-rs", branch = "main" }
-fork = "0.1"
 
 [profile.release]
 opt-level = "z"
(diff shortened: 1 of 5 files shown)


madninja (commented):
1. Clippy left some warnings for you to fix up. Run with `cargo +nightly  clippy -- -Dclippy::all ` to reproduce locally
2. Could you try using daemonize just like main.rs does? I believe it does the same thing as the fork crate without having to introduce another crate
3. I _think_ the logger issue may require you to re-setup the logger after the fork/daemon run, but I'm not sure sure

Otherwise this will be awesome!

PaulVMo:
Thanks for the feedback. I will address clippy and work on logging. 

Regarding switching to daemonize, calling start() kills the parent I believe which is no good in this case since we only want the parent to be terminated by the actual opkg install running.

There is an unreleased feature in daemonize, execute(), which leaves the parent running. It is not in the latest crate. Is it still possible to use that?
https://github.com/knsd/daemonize/commit/e1546c35a4638721b73dcbe6f162e9b24c5d7669

madninja:
> Thanks for the feedback. I will address clippy and work on logging.
> 
> Regarding switching to daemonize, calling start() kills the parent I believe which is no good in this case since we only want the parent to be terminated by the actual opkg install running.
> 
> There is an unreleased feature in daemonize, execute(), which leaves the parent running. It is not in the latest crate. Is it still possible to use that? [knsd/daemonize@e1546c3](https://github.com/knsd/daemonize/commit/e1546c35a4638721b73dcbe6f162e9b24c5d7669)

Yeah we can switch to the non crates release reference and ask them for a crates.io push?

Cargo.toml
```diff
@@ -38,7 +38,7 @@ thiserror = "1.0"
 base64 = "0"
 rand = "0.8"
 prost = "0"
-daemonize = "0.4"
+daemonize = {git = "https://github.com/knsd/daemonize", rev = "f7be28e"}
```
madninja, line 41: I didn't think we'd need a specific rev?

madninja (commented):
One question and some clippy warnings to address

Run ```cargo +nightly  clippy -- -Dclippy::all ```

Cargo.toml
PaulVMo: We need something newer than the latest release which was tagged over a year ago. I was debating using the main branch but thought that opened things up for  unexpected changes.

PaulVMo:
Thanks. I'll take care of clippy but also might need some more help or suggestions on the logging. The logger in the daemon process still doesn't work. Sorry should have left a comment that I was working on that still

madninja:
> Thanks. I'll take care of clippy but also might need some more help or suggestions on the logging. The logger in the daemon process still doesn't work. Sorry should have left a comment that I was working on that still

unclear on what the right path is there.. I _think_ maybe you'd have to recreate logger instance in the new daemon?

madninja:
WIth #339 this is no longer needed

PR #262, opened 2022-06-12
Adjust tx_power with asserted gain

madninja:


src/router/client.rs
```diff
@@ -495,7 +496,7 @@ impl RouterClient {
         StateChannelMessage::offer(
             packet.packet().clone(),
             self.keypair.clone(),
-            self.region,
+            &self.region,
```
JayKickliter, line 499: Minor nit: region isn’t Copy, so why this change?

madninja: Because the offer call only needs a reference to it to construct the representing internal value

src/region.rs
```diff
@@ -99,3 +101,56 @@ impl Region {
...
+        Ok(Self {
+            gain: value.gain,
+            params,
+            region,
+        })
+    }
+}
+
+impl RegionParams {
+    pub fn max_eirp(&self) -> Option<u32> {
```
JayKickliter, line 141: A comment on units would be helpful. Deci-dBm?

JayKickliter: Or maybe even a fixed-point type? I'd prefer to reduce the risk of someone down the line not taking into account a scaling factor.

madninja: Added in [63f3de5](https://github.com/helium/gateway-rs/pull/262/commits/63f3de5bcdcd9e7cfdf0856edae28eabc98f2487)

JayKickliter (approved):

Issue #264, opened 2022-06-18
Gateway resets region even when specified


bsodmike:
Hi,

I noticed that even though I have the region specified in my settings.toml, when the gateway service is restarted it will fetch its "region" based on its location. 

```
root@RAK7268C:/etc/helium_gateway# helium_gateway info
{
  "fw": "ramips_24kec-1.0.0-alpha.27",
  "region": "AS923_1",
```

Settings toml
```
root@RAK7268C:/etc/helium_gateway# cat settings.toml
listen = "0.0.0.0:1680"
region = "EU868"

[log]
method = "stdio"
level = "info"
timestamp = false

[update]
platform = "ramips_24kec"
```

You can see how it overrides this here:

```
Sat Jun 18 08:24:12 2022 daemon.info helium_gateway[22936]:  INFO starting, module: updater
Sat Jun 18 08:24:12 2022 daemon.info helium_gateway[22936]:  INFO starting, listen: 127.0.0.1:4467, module: api
Sat Jun 18 08:24:12 2022 daemon.info helium_gateway[22936]:  INFO new packet forwarder client: AC:1F:09:FF:FE:06:14:FA, 127.0.0.1:59021, module: g
ateway
Sat Jun 18 08:24:13 2022 daemon.info helium_gateway[22936]:  INFO using gateway, uri: http://205.209.96.147:8080/, pubkey: 11qcKRFmJuPEnxcgWoP7Yvk
zKr1dCrp1M9MCBjaCQGmgexKtcCZ, module: dispatcher
Sat Jun 18 08:24:15 2022 daemon.info helium_gateway[22936]:  INFO checking gateway, block_age: 574, pubkey: 11qcKRFmJuPEnxcgWoP7YvkzKr1dCrp1M9MCBj
aCQGmgexKtcCZ, module: dispatcher
Sat Jun 18 08:24:15 2022 kern.info quectel-CM[21119]: requestRegistrationState2 MCC: 413, MNC: 8, PS: Detached, DataCap: UNKNOW
Sat Jun 18 08:24:15 2022 kern.info quectel-CM[21119]: Dail Failed 3
Sat Jun 18 08:24:15 2022 daemon.info helium_gateway[22936]:  INFO updated region to AS923_1 at height 1402113, module: dispatcher
```

IoTThinks:
By right, the region in location_assert should override the region in setting.toml

https://github.com/helium/gateway-rs/releases
[...]

PR #265, opened 2022-06-18
Allow automatic region updates to be disabled via settings

bsodmike:
This is a proposed patch for https://github.com/helium/gateway-rs/issues/264

### DIFF
--- a/config/default.toml
+++ b/config/default.toml
@@ -20,6 +20,8 @@ keypair = "/etc/helium_gateway/gateway_key.bin"
 listen = "127.0.0.1:1680"
 api = 4467
 region = "US915"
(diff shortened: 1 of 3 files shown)


madninja:
This is not ideal since you have no idea when the user is updating the hotspot location on chain. Why do you need this?

madninja:
I see the issue comment. think this should be inverted where the specified settings region is used as a fallback if the user hasn’t asserted location. 

Note that ends up disabling downlinks since the region parameters are required for inferring tx power

bsodmike:
Closed due to advice provided in https://github.com/helium/gateway-rs/issues/264

PR #271, opened 2022-07-11
send packets as quickly as possible

michaeldjeffrey:
could it be this easy?


src/router/client.rs
```diff
@@ -232,14 +228,11 @@ impl RouterClient {
...
-        if self.first_uplink {
-            self.first_uplink = false;
-            self.maybe_connect_state_channel(logger).await;
-        }
-        self.send_packet_offers(logger).await
+        // REVIEW: There was previously logic to handle first connection attempts.
+        // Is that still necessary, or will this just send packets into the void
+        // until a good connection can be made?
+        let packet = QuePacket::from(uplink);
+        self.send_packet(logger, Some(&packet)).await
```
madninja, line 235: this isn't quite right.. you'd want to still queue packets if you can't reach router.. and then have a _send_packets_ that sends them all when it connects?

PR #289, opened 2022-08-10
Packet Router Client

michaeldjeffrey:
Removes starting routers for OUIs, only starts connections for default routers, and sends all packets to them.

- Going from `PacketRouterPacketDownV1` to `Packet`, I'm not sure what the proper handling of the rx windows should be. 
- I wanted to go from `RxPk` to `PacketRouterPacketUpV1` as soon as possible, but that would mean leaving some important fields empty until the last moment (`hold_time` and the likes), and I'm hoping there's a smarter way so we don't have to rely on memory.


src/router/client.rs
```diff
@@ -150,32 +156,18 @@ impl RouterClient {
...
-            self.keypair.clone(),
-            &self.region,
-            packet.hold_time().as_millis() as u64,
-        )
-        .and_then(|message| self.router.route(message.to_message()))
-        .map_ok(StateChannelMessage::from_message)
-        .await
+
+        packet
+            .to_packet_up(self.keypair.clone(), &self.region)
```
madninja, line 169: rename to `to_uplink` if you could?

src/router/dispatcher.rs
```diff
@@ -277,7 +279,7 @@ impl Dispatcher {
                 },
                 gateway_message = streams.next() => match gateway_message {
                     Some((gateway_stream, Ok(gateway_message))) => match gateway_stream {
-                        GatewayStream::Routing => self.handle_routing_update(&gateway_message, &shutdown, logger).await,
+                        GatewayStream::Routing => info!(logger, "ignoring routing update"),
```
madninja, line 282: remove this by removing it from the `GatewayStream` enum and don't subscribe to routing updates

src/router/routing.rs
```diff
@@ -1,43 +1,18 @@
-use super::{DevAddrFilter, EuiFilter};
 use crate::{KeyedUri, PublicKey, Result};
-use helium_proto::{routing_information::Data as RoutingData, RoutingInformation};
 use slog::{warn, Logger};
 use std::{convert::TryFrom, sync::Arc};
 
 #[derive(Clone, Debug)]
 pub struct Routing {
```
madninja, line 6: I think we need to get rid of Routing altogether and have routers be a DNS name(s) for regional delivery right? this would remove the public key of those routers which will be something we need to talk about.

src/traits/msg_sign.rs
```diff
@@ -36,6 +37,7 @@ impl_msg_sign!(GatewayRegionParamsUpdateReqV1, signature);
 impl_msg_sign!(BlockchainStateChannelPacketV1, signature);
 impl_msg_sign!(BlockchainStateChannelOfferV1, signature);
 impl_msg_sign!(BlockchainTxnStateChannelCloseV1, signature);
+impl_msg_sign!(PacketRouterPacketUpV1, signature);
```
madninja, line 40: can we rename the proto to `PacketRouterUplinkV1` ?

src/router/routing.rs
michaeldjeffrey: Routing has been removed. 
I'm unsure of the implications of removing the router public keys. But like you said, that's something we can talk about.

src/traits/msg_sign.rs
michaeldjeffrey: I do like that name. @macpie thoughts?

macpie: I like the `v1` but not `uplink` because not only `uplinks` come threw there but `joins` as well

PR #299, opened 2022-09-19
use proto crate with protoc vendored

madninja:


madninja:
wrong branch commit for PR

PR #301, opened 2022-09-23
implement beacon transmits

madninja:


beacon/src/beacon.rs
```diff
@@ -0,0 +1,84 @@
...
+        };
+        let mut seed = [0u8; 32];
+        seed.copy_from_slice(&data[0..32]);
+        let mut rng = rand_chacha::ChaCha12Rng::from_seed(seed);
+
+        let frequency = rand_frequency(region_params, &mut rng);
+        let datarate = DataRate::Sf7bw125;
+
+        Self {
+            data: data[0..BEACON_PAYLOAD_SIZE].to_vec(),
```
JayKickliter, line 42: You already have a Vec that's about to drop, might as well use it
```suggestion
            data: {
                data.truncate(BEACON_PAYLOAD_SIZE);
                data
            },
```

beacon/src/beacon.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn rand_frequency<R>(region_params: &[BlockchainRegionParamV1], rng: &mut R) -> u64
+where
+    R: Rng + ?Sized,
+{
+    assert!(!region_params.is_empty());
+    region_params
+        .choose(rng)
+        .map(|params| params.channel_frequency)
+        .unwrap()
```
JayKickliter, line 63: Since `assert!` is live for release bulds. You get get rid of it and use expect below
```suggestion
    region_params
        .choose(rng)
        .map(|params| params.channel_frequency)
        .expect("empty regional parameters")
```

beacon/src/beacon.rs
```diff
@@ -0,0 +1,84 @@
...
+            local_entropy: v.local_entropy.data,
+            remote_entropy: v.remote_entropy.data,
+            data: v.data,
+            frequency: v.frequency,
+            channel: 0,
+            datarate: v.datarate as i32,
+            tx_power: 27,
+            timestamp: SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_else(|_| Duration::from_secs(0))
```
JayKickliter, line 79: is this a place holder to simplify the code, or meant for real world use? If temporary, a `// TODO: ...` is warranted.

beacon/src/entropy.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+impl Entropy {
+    pub fn local() -> Self {
+        let mut local_entropy = [0u8; LOCAL_ENTROPY_SIZE];
+        OsRng.fill_bytes(&mut local_entropy);
+        let version = default_version();
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_else(|_| Duration::from_secs(0))
```
JayKickliter, line 24: same as earlier comment about perhaps adding a `// TODO` comment here

beacon/src/entropy.rs
```diff
@@ -0,0 +1,103 @@
...
+            .as_secs() as i64;
+        Self {
+            version,
+            timestamp,
+            data: local_entropy.to_vec(),
```
JayKickliter, line 29: Might as well keep the Vec and skip the intermediate stack allocation

```suggestion
        let mut local_entropy = vec![0u8; LOCAL_ENTROPY_SIZE];
        OsRng.fill_bytes(&mut local_entropy);
        let version = default_version();
        let timestamp = SystemTime::now()
            .duration_since(UNIX_EPOCH)
            .unwrap_or_else(|_| Duration::from_secs(0))
            .as_secs() as i64;
        Self {
            version,
            timestamp,
            data: local_entropy,
```

beacon/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[dependencies]
+rand_chacha = "0"
+helium-proto = {workspace = true}
+rand = {workspace = true}
+base64 = {workspace = true}
+serde = {workspace = true}
+sha2 = {workspace = true}
+
+[dev-dependencies]
+serde_json = "1"
```
JayKickliter, line 17:
```suggestion
serde_json = "1"

```

src/api/server.rs
```diff
@@ -136,7 +136,7 @@ impl Api for LocalServer {
             .ecdh(&public_key)
             .map_err(|_err| Status::internal("Failed ecdh"))?;
         let reply = EcdhRes {
-            secret: secret.as_bytes().to_vec(),
+            secret: secret.raw_secret_bytes().to_vec(),
```
JayKickliter, line 139: was this a bugfix?

src/region.rs
```diff
@@ -170,6 +143,13 @@ impl RegionParams {
             .and_then(|max_eirp| (max_eirp - self.gain).trunc().to_u32())
     }
 
+    pub fn rand_frequency(&self) -> Option<u64> {
+        self.params
+            .as_slice()
+            .choose(&mut OsRng)
+            .map(|params| params.channel_frequency)
+    }
+
```
JayKickliter, line 152: Not needed.

```suggestion
```

JayKickliter (approved):
No showstoppers, good as is. All my suggested changes can be filed under "best practice, but unnecessary"

beacon/src/beacon.rs
madninja: It's for real use.. i.e. the timestamp in the report is set to when it is created. I'll add a comment

JayKickliter: I was specifically referring to `.unwrap_or_else(|_| Duration::from_secs(0))`, which will be interpreted by the receiver of this message to be the unix epoch itself.

lorawan/src/lib.rs
```diff
@@ -64,6 +87,18 @@ const JOIN_ACCEPT_WITH_CFLIST_LEN: usize = 33;
 const DATA_MIN_LEN: usize = 12;
 
 impl PHYPayload {
+    pub fn propietary(payload: &[u8]) -> Self {
```
Vagabond, line 90: proprietary

JayKickliter (approved):
Other than me spelling "proprietary" wrong, all good. It's your choice on whether or not that should be fixed.

beacon/src/beacon.rs
```diff
@@ -20,34 +21,53 @@ pub struct Beacon {
...
-                data
-            },
-            frequency,
-            datarate,
-            local_entropy,
-            remote_entropy,
-        })
+                let frequency = rand_frequency(region_params, &mut rng)?;
+                let payload_size =
+                    rng.gen_range(MIN_BEACON_V0_PAYLOAD_SIZE..MAX_BEACON_V0_PAYLOAD_SIZE);
```
JayKickliter, line 55: Unless gen_range does something special, it'll never pick `MAX_BEACON_V0_PAYLOAD_SIZE` since ranges in rust are not inclusive.

madninja: you're right.. fixing

beacon/src/beacon.rs
```diff
@@ -20,34 +21,53 @@ pub struct Beacon {
...
-            let mut hasher = Sha256::new();
-            remote_entropy.digest(&mut hasher);
-            local_entropy.digest(&mut hasher);
-            hasher.finalize().to_vec()
-        };
-        let mut seed = [0u8; 32];
-        seed.copy_from_slice(&data[0..32]);
-        let mut rng = rand_chacha::ChaCha12Rng::from_seed(seed);
+        if remote_entropy.version != local_entropy.version {
+            return Err(Error::invalid_version());
```
Vagabond, line 38: I'm not sure about this check, what happens if the code support entropy 0 and 1, and we switch it on the server, the client cannot allow both versions to exist across the update?

madninja: I'll just drop local version then.. It's really not relevant what the version is

Issue #298, opened 2022-09-13
show "version `GLIBC_2.29' not found" after Release v1.0.0-alpha.29


andrewintw:
I am using Debian GNU/Linux 10 (buster). When I install helium-gateway-v1.0.0-alpha.27-raspi_64.deb and helium-gateway-v1.0.0-alpha.28-raspi_64.deb. It works well. But after helium-gateway-v1.0.0-alpha.29-raspi_64.deb, it shows 

```
helium_gateway: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by helium_gateway)
```

Even if I rollback to v1.0.0-alpha.27, it will update to 1.0.0~alpha.31 automatically and show the same error message.

I don't know why you upgrade the dependency clib version, that is a big change and not a good choice for an embedded system.

madninja:
It's part of a build system upgrade, not intentional. There's no good way to roll that back afaict, but if you know a way I'd love to know

andrewintw:
> It's part of a build system upgrade, not intentional. There's no good way to roll that back afaict, but if you know a way I'd love to know

In general, Helium should maintain a fixed libc version if there are no specific reason. The original design idea is that device to update automatically, but automatic update should not force the device to upgrade the whole system. This is unreasonable for embedded systems.

The best solution is to roll back the old build environment, or use static-linking to generate the executable. Otherwise, any device will not be able to run helium_gateway after updated via Helium server.

madninja:
> > It's part of a build system upgrade, not intentional. There's no good way to roll that back afaict, but if you know a way I'd love to know
> 
> In general, Helium should maintain a fixed libc version if there are no specific reason. Once libc is upgraded, it usually means a system-wide upgrade. I can imagine that Helium release .deb for different platforms so that those platforms can be updated automatically. But automatic update should not force the device to upgrade the whole system. This is unreasonable for embedded systems.
> 
> The best solution is to roll back the old build environment, or use static-linking to generate the executable. Otherwise, any device will not be able to run helium_gateway after updated via Helium server.
> 
> If Helium still wants to keep the new build system and use the same build process, then I think all platform maintainers will be forced to stop using automatic upgrades and build binaries from source. They also need to design their own upgrade procedure.


I understand all the arguments.. If we don't want to hold back toolchain upgrades which really help with performance and stability, what are our options? I'm looking for help, not reasons to go to the past

andrewintw:
[...]

PR #302, opened 2022-09-27
static linking

qbaj:
Setting static linking for all targets in project due to https://github.com/helium/gateway-rs/issues/298

Musl libc is static cross and native toolchains, so targets with *-linux-musl should be static linking.
But not all, there are some static and dynamic linking (more info eg. https://github.com/rust-lang/rust/issues/80693),
so flag static linking has been set for all.

Exception for x86_64-unknown-linux-gnu (targets: x86_64 and freedomfi), there are some linking problem:
```
warning: Using '*' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```
and it should be investigated.

tasks.info prints linking dependencies for packages.


madninja:
thanks @qbaj ! The main concern I have with this PR is that it affects _all_ targets which you are not going to be able to test. I know, for example, that at least on one or two of those targets a static build would make the binary to large to fit in the writable part of the filesystem. 

Could you make this specific to just the one or two targets we know have issues?

qbaj:
Could you give these targets which have a problem with to large binary?

I can prepare config for targets with issues like raspi_64 - https://github.com/helium/gateway-rs/issues/298,
but I think all targets should have it, especially embedded targets where os upgrade is difficult.

I checked all targets from Supported Platforms and we have:
- static linking:
  - mipsel-unknown-linux-musl
  - armv5te-unknown-linux-musleabi
  - armv7-unknown-linux-musleabihf
- dynamic linking:
  - mips-unknown-linux-musl
  - x86_64-unknown-linux-gnu
  - aarch64-unknown-linux-gnu
  - arm-unknown-linux-gnueabihf
  - armv7-unknown-linux-gnueabihf

So all targets with -musl are static (musl is lightweight design of C standard library to static linking), except mips and mipsel (https://github.com/rust-lang/rust/issues/80693), but mipsel is different and in our system is static too. 
All targets with -gnu and mips/musl could have problem with linking newer glibc at this time. 

Also I checked sizes of binaries:
| target | static | dynamic |
| - | - | - |
| mips-unknown-linux-musl | 4519268 | 4444720 | 
| aarch64-unknown-linux-gnu | (link error ??) | 3356400 |
| arm-unknown-linux-gnueabihf | 3655752 | 3186008 | 
| armv7-unknown-linux-gnueabihf | 3456016 | 3145228 |

To consider: static linking for all targets and/or switch gnu to musl in all targets.

PR #304, opened 2022-10-07
Distinguish between failure to execute and failed execution

xandkar:


src/curl.rs
```diff
@@ -17,8 +17,18 @@ where
         .arg(&url)
         .output()
         .map(move |result| match result {
-            Ok(output) => f(&output.stdout),
-            Err(err) => Err(Error::from(err)),
+            Ok(output) => {
+                if output.status.success() {
```
madninja, line 21: Can you use more idiomatic rust case if expression here?

also not quite a fan of the custom error. But it does give more context than before

src/curl.rs
```diff
@@ -17,8 +18,23 @@ where
...
+                Some(n) => Err(Error::Custom(format!(
+                    "curl exited with non-0 status: {}. stderr: {:?}",
+                    n, stderr
+                ))),
+                None => Err(Error::Custom(format!(
+                    "curl terminated by signal. stderr: {:?}",
+                    stderr
+                ))),
+            },
+            Err(err) => Err(Error::Custom(format!("curl failed to execute: {:?}", err))),
```
xandkar, line 37: @madninja Not sure what exactly you meant by "more idiomatic rust case if expression", but I switched it to match on status code, which uncovered yet another failure variety - termination by signal.

Stringy errors _are_ kinda smelly, but IMO fine for errors you intend to just log rather than case-analyze. What would you suggest instead?

xandkar: Maybe define something like `Error::Cmd`?

madninja: Since the curl module itself isn't the top level app, perhaps we define a this_error derived Error for curl errors? That contains the curl errors and makes them more contextual. Then have gateway-rs's Error enum enhanced with a `CurlError(curl::Error)`

jeffgrunewald: As for the “idiomatic match casing” I would lean toward a railway style ` if let` syntax for the success case and catch the failed exit code or failed execution in the `else` blocks

PR #308, opened 2022-10-07
fixing workspace dependencies; config values

jeffgrunewald:
Fixes some configuration values for the default endpoints the helium_gateway should call to in order to receive the remote entropy data as well as the default (mainnet) endpoint for submitting beacon reports to.

re-introduces the tonic client request interceptor from early testing to allow injecting an "auth" token into the post requests to the grpc beacon ingestor service.

since these requests will be over plaintext http2 and not https2 this token is intended more as a "fail fast if the beacon is not intended for the network in question and as a result is named "ingest token" instead of "auth token" (or similar) in the public-facing configuration and defaults to `mainnet`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ log = "0"
 bytes = "*"
 xxhash-rust = { version = "0.8", features = ["xxh64"]}
 xorf = "0.7"
-sha2 = {workspace = true }
+sha2 = {workspace = true}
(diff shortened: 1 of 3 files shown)


jeffgrunewald:
Not sure what to make of this now; I have both successfully and _not_ successfully built the project with the main dependencies inherited from the `workspace.dependencies` key.

bbalser (approved):

PR #312, opened 2022-10-11
Derive Max Transmit Power from Tx Error

lthiery:
Whenever we transmit, we spawn a thread and await the ack which may contain a tx power error. I extended the existing channel for clients in the gateway module so that I could send the maximum tx power from a spawned thread.

### DIFF
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -73,7 +73,7 @@ pub struct Gateway {
     listen_address: String,
     region_params: Option<RegionParams>,
     /// If packet forwarder returns InvalidTxPower, it provides the max allowable tx value
-    /// which we can save and apply it as a maximum to outgoing packets
+    /// which we can save and apply as a maximum to future outgoing packets
     max_power: Option<u32>,
 }
 
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -22,6 +22,7 @@ pub enum Message {
     Downlink(Packet),
     TransmitBeacon(Beacon),
     RegionParamsChanged(RegionParams),
+    MaxPower(u32),
 }
 
 #[derive(Clone, Debug)]
@@ -64,17 +65,22 @@ impl MessageSender {
 
 pub struct Gateway {
     uplinks: dispatcher::MessageSender,
+    self_sender: MessageSender,
     messages: MessageReceiver,
     beacon_handler: beaconer::MessageSender,
     downlink_mac: MacAddress,
     udp_runtime: UdpRuntime,
     listen_address: String,
     region_params: Option<RegionParams>,
+    /// If packet forwarder returns InvalidTxPower, it provides the max allowable tx value
+    /// which we can save and apply it as a maximum to outgoing packets
+    max_power: Option<u32>,
 }
 
 impl Gateway {
     pub async fn new(
         uplinks: dispatcher::MessageSender,
+        self_sender: MessageSender,
         messages: MessageReceiver,
         beacon_handler: beaconer::MessageSender,
         settings: &Settings,
@@ -82,11 +88,13 @@ impl Gateway {
         let gateway = Gateway {
             uplinks,
             downlink_mac: Default::default(),
+            self_sender,
             messages,
             beacon_handler,
             listen_address: settings.listen.clone(),
             udp_runtime: UdpRuntime::new(&settings.listen).await?,
             region_params: None,
+            max_power: None,
         };
         Ok(gateway)
     }
@@ -160,6 +168,7 @@ impl Gateway {
 
     async fn handle_message(&mut self, logger: &Logger, message: Message) {
         match message {
+            Message::MaxPower(max_power) => self.max_power = Some(max_power),
             Message::Downlink(packet) => self.handle_downlink(logger, packet).await,
             Message::TransmitBeacon(beacon) => self.handle_transmit_beacon(logger, beacon).await,
             Message::RegionParamsChanged(region_params) => {
@@ -182,7 +191,11 @@ impl Gateway {
         };
 
         if let Some(tx_power) = region_params.tx_power() {
-            Some(tx_power)
+            Some(if let Some(max_power) = self.max_power {
+                std::cmp::min(max_power, tx_power)
+            } else {
+                tx_power
+            })
         } else {
             warn!(logger, "ignoring beacon transmit, no tx power");
             None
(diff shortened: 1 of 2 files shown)


src/gateway.rs
```diff
@@ -173,18 +182,30 @@ impl Gateway {
...
             region_params
         } else {
             warn!(logger, "ignoring transmit request, no region params");
-            return;
+            return None;
         };
 
-        let tx_power = if let Some(tx_power) = region_params.tx_power() {
-            tx_power
+        if let Some(tx_power) = region_params.tx_power() {
```
madninja, line 193: surely there's a clearer way to do all this Option wrangling?

src/gateway.rs
```diff
@@ -173,18 +182,30 @@ impl Gateway {
         }
     }
 
-    async fn handle_transmit_beacon(&mut self, logger: &Logger, beacon: Beacon) {
+    fn tx_power(&mut self, logger: &Logger) -> Option<u32> {
         let region_params = if let Some(region_params) = &self.region_params {
             region_params
         } else {
             warn!(logger, "ignoring transmit request, no region params");
```
madninja, line 189: I'd prefer tx_power just return a Result<u32> instead of an Option<u32> and not log in something that's just there to return a value? i.e. do the logging in the caller(s)?

src/gateway.rs
```diff
@@ -64,29 +65,36 @@ impl MessageSender {
...
     uplinks: dispatcher::MessageSender,
+    self_sender: MessageSender,
     messages: MessageReceiver,
     beacon_handler: beaconer::MessageSender,
     downlink_mac: MacAddress,
     udp_runtime: UdpRuntime,
     listen_address: String,
     region_params: Option<RegionParams>,
+    /// If packet forwarder returns InvalidTxPower, it provides the max allowable tx value
+    /// which we can save and apply it as a maximum to outgoing packets
```
madninja, line 76: so the invalid tx value is returned but the packet is still sent with that indicated tx power  right?

src/gateway.rs
```diff
@@ -64,29 +65,36 @@ impl MessageSender {
...
+    max_power: Option<u32>,
 }
 
 impl Gateway {
     pub async fn new(
         uplinks: dispatcher::MessageSender,
+        self_sender: MessageSender,
```
madninja, line 83: this seems like a hack? you're passing in a MessageSender that assumes it's a sender for the given messages receiver

src/gateway.rs
```diff
@@ -207,32 +229,35 @@ impl Gateway {
             {
                 Ok(()) => info!(logger, "beacon transmitted"; "beacon" => &beacon_id),
                 Err(err) => {
+                    if let semtech_udp::server_runtime::Error::Ack(
+                        tx_ack::Error::InvalidTransmitPower(Some(value)),
+                    ) = err
+                    {
+                        if let Err(e) = self_sender.send(Message::MaxPower(value as u32)).await {
```
madninja, line 236: use inspect_err to do the warn (it's in FutureExt iirc)

src/gateway.rs
jeffgrunewald: is it considered "correct" if there's no tx_power? either way we're sort of bailing out on sending the beacon in the caller so seems like you could just as easily have the final value returned as a result?

madninja: My point was that the tx_power function is logging on behalf of the caller and ends up logging, for example, about "beacon" even when it's being used for a downlink.

jeffgrunewald: Two parts to that statement; I agree I wouldn’t have this function do the logging but to the other part of your comment I’m wondering if it’s appropriate for this function to return an option or a result

lthiery: We are logging from a `fn(&self)` - seems innocuous to me; error msg does need fixing though.

If we make this a result type, each `tx_power` caller (`handle_beacon`, `handle_downlink`) will have to unpack Result, draft an error message, etc. Seems tedious when the point of refactoring this into a function was to reduce the repetition between those two functions.

src/gateway.rs
lthiery: Fundamentally, there's two ways of solving the original problem, ie: "How to send InvalidTransmitPower back after being spawned off"

Solution 1: Mutex
Solution 2: Channel

I think interacting directly with mutexes is annoying, so I'd rather extend the use of an existing channel. And I don't think sending messages to my own queue is a hack.

src/gateway.rs
lthiery: I tried but it looks like inspect_err wants `FnOnce(&Self::Error)` which is not async. I can't handle a mutex or a channel if I'm not in an async code block.

src/gateway.rs
lthiery: Yes, you're right. Just pushed something that I think works better.

src/gateway.rs
lthiery: The InvalidTxValue originally stems from the packet forwarder which will NACK a packet with power that is too high; the frame is dropped by the packet forwarder.

In that NACK, however, it provides the max allowable power. After that, gateway-rs will never trigger the error again from packet forwarder as it saves the max_power value and applies to to future packets; it does this quietly within `fn tx_power` without alerting the user.

src/gateway.rs
```diff
@@ -173,18 +182,28 @@ impl Gateway {
...
             region_params
         } else {
-            warn!(logger, "ignoring transmit request, no region params");
-            return;
+            warn!(logger, "ignoring transmit: no region params");
+            return None;
         };
 
-        let tx_power = if let Some(tx_power) = region_params.tx_power() {
+        if let (Some(tx_power), Some(max_power)) = (region_params.tx_power(), self.max_power) {
```
jeffgrunewald, line 193: what about combining the above into something like:
```
if let Some(region_params) = &self.region_params {
    match (region_params.tx_power(), self.max_power) {
        (Some(tx_power), Some(max_power)) => Some(std::cmp::min(tx_power, max_power)),
        (Some(tx_power), _) => Some(tx_power),
        _ => None
    }
}
```

lthiery:
After discussions with @ke6jjj , I'm realizing that TX_POWER warning from packet forwarder is only a warning. The packet has still sent out.

lthiery:
Closing this PR in favor or something else

src/gateway.rs
madninja: right but at least they _can_ provide contextual logging

src/gateway.rs
madninja: oh man it's dropped by the packet forwarder? I was convinced it would return what power it transmitted at

src/gateway.rs
madninja: _OR_ use another message channel instead of assuming that the self_sender will send to messages

src/gateway.rs
madninja: huh? self_sender.send is a figure right? so inspect_err would work before the await?

PR #313, opened 2022-10-11
Avoid reusing ingest connections by creating a new one each time

Vagabond:


jeffgrunewald (commented):
did we also want this to retry on failure for any amount/number of times?

assuming tonic automatically does the right thing when these ephemeral client's are dropped and closes all the connections we're trying to stop reusing.

Vagabond:
we certainly can do a retry, but this was more about preventing the reuse problem to begin with, which it seems to do in testing.

jeffgrunewald (approved):

madninja (approved):
This looks good.. Ideally there'd be a corresponding issue to request queued beacon and witness reports?

jeffgrunewald:
> This looks good.. Ideally there'd be a corresponding issue to request queued beacon and witness reports?

You mean a GitHub issue to add this feature to the local grpc api?

madninja:
> > This looks good.. Ideally there'd be a corresponding issue to request queued beacon and witness reports?
> 
> You mean a GitHub issue to add this feature to the local grpc api?

No sorry that was confusing.. an issue, but an issue to add _queued_ witness/beacon report submission in case the hotspot fails to deliver

jeffgrunewald:
> > > This looks good.. Ideally there'd be a corresponding issue to request queued beacon and witness reports?
> > 
> > 
> > You mean a GitHub issue to add this feature to the local grpc api?
> 
> No sorry that was confusing.. an issue, but an issue to add _queued_ witness/beacon report submission in case the hotspot fails to submit the beacon or witness report

Why don’t we take 10 min tomorrow to figure out what the parameters for this are and I’ll write it up

PR #320, opened 2022-10-14
Hoist witness report signing into beacon handling function

Vagabond:


src/beaconer.rs
```diff
@@ -140,16 +140,19 @@ impl Beaconer {
...
+        let mut report = match packet.to_witness_report() {
             Ok(report) => report,
             Err(err) => {
                 warn!(logger, "ignoring invalid witness report: {err:?}");
                 return;
             }
         };
 
+        report.pub_key = self.keypair.public_key().to_vec();
+        report.signature = report.sign(self.keypair.clone()).await.unwrap();
```
jeffgrunewald, line 152:
```suggestion
        report.signature = report.sign(self.keypair.clone()).await?;
```

jeffgrunewald: the raw unwrap here will cause a panic whereas the enigmatic `?` will just throw an error up the stack

madninja: indeed, unwraps continue to be considered evil

madninja (approved):

PR #322, opened 2022-10-18
Secondary beacons

Vagabond:
To allow for more packet events on the same timebase, implement a random chance that witnesses will transmit a secondary beacon.

### DIFF
--- a/beacon/src/entropy.rs
+++ b/beacon/src/entropy.rs
@@ -42,7 +42,7 @@ impl Entropy {
             .map_err(Error::from)?
             .as_secs() as i64;
         Ok(Self {
-            version: 0,
+            version: 1, // marked as local
             timestamp,
             data: data,
         })
--- a/beacon/src/entropy.rs
+++ b/beacon/src/entropy.rs
@@ -36,6 +36,18 @@ impl Entropy {
         })
     }
 
+    pub fn from_data(data: Vec<u8>) -> Result<Self> {
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map_err(Error::from)?
+            .as_secs() as i64;
+        Ok(Self {
+            version: 0,
+            timestamp,
+            data: data,
+        })
+    }
+
     pub(crate) fn digest<D: Digest>(&self, state: &mut D) {
         state.update(&self.data);
         state.update(self.timestamp.to_ne_bytes());
--- a/beacon/src/entropy.rs
+++ b/beacon/src/entropy.rs
@@ -44,7 +44,7 @@ impl Entropy {
         Ok(Self {
             version: 1, // marked as local
             timestamp,
-            data: data,
+            data,
         })
     }
 
--- a/src/beaconer.rs
+++ b/src/beaconer.rs
@@ -11,14 +11,13 @@ use crate::{
     sync, Base64, Error, Keypair, MsgSign, Packet, RegionParams, Result,
 };
 use futures::TryFutureExt;
-use helium_proto::services::poc_lora;
-use helium_proto::Message as OtherMessage;
+use helium_proto::{services::poc_lora, Message as ProtoMessage};
 use http::Uri;
 use slog::{self, info, warn, Logger};
 use std::{sync::Arc, time::Duration};
 use tokio::time;
 use triggered::Listener;
-use xxhash_rust::xxh64::Xxh64;
+use xxhash_rust::xxh64::xxh64;
 
 /// Message types that can be sent to `Beaconer`'s inbox.
 #[derive(Debug)]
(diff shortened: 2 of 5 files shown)


src/beaconer.rs
```diff
@@ -164,6 +168,58 @@ impl Beaconer {
             }
         };
 
+        // check if hash of witness is below the "difficulty threshold" for a secondary beacon
+        let buf = Sha256::digest(report.encode_to_vec()).to_vec();
+        let threshold = BigInt::parse_bytes(
+            b"11388830933659919894162346859235831137017100287433801699915283717462644789984",
+            10,
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
Vagabond: My hope for this was actually to get the value from the entropy server so we could adjust the "rate" of secondary beacons over time.

src/beaconer.rs
```diff
@@ -164,6 +168,58 @@ impl Beaconer {
...
+                report.tmst = tmst;
+
+                let _ = PocLoraService::new(self.poc_ingest_uri.clone())
+                        .submit_beacon(report, self.keypair.clone())
+                        .inspect_err(|err| info!(logger, "failed to submit secondary poc beacon report: {err:?}"; "beacon" => &beacon_id))
+                        .inspect_ok(|_| info!(logger, "poc secondary beacon report submitted"; "beacon" => &beacon_id))
+                        .await;
+            }
+        } else {
+            info!(
```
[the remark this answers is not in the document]
madninja: I agree.. just got it as soon as I saw this part of the branch

src/beaconer.rs
```diff
@@ -164,6 +168,83 @@ impl Beaconer {
...
+
+            let _ = PocLoraService::new(self.poc_ingest_uri.clone())
+                .submit_beacon(report, self.keypair.clone())
+                .inspect_err(|err| info!(logger, "failed to submit secondary poc beacon report: {err:?}"; "beacon" => &beacon_id))
+                .inspect_ok(|_| info!(logger, "poc secondary beacon report submitted"; "beacon" => &beacon_id))
+                .await;
+        } else {
+            info!(
+                logger,
+                "no secondary beacon, threshold {threshold:?}, factor {factor:?}"
```
madninja, line 244: make this slog attributes instead of inlines in the string.. see how the "beacon" => bacon_id is used above

src/beaconer.rs
```diff
@@ -164,6 +168,83 @@ impl Beaconer {
...
+            10,
+        )
+        .unwrap();
+        // compare the hash of the witness report as a bigint to the difficulty threshold
+        // this is sort of a bitcoin-esque proof of work check insofar as as we're looking
+        // for hashes under a certain value. Because of the time constraints involved this
+        // should not be a 'mineable' check, but it provides a useful probabalistic way to
+        // allow for verifiable secondary beacons without any coordination.
+        let factor = BigInt::from_bytes_be(Sign::Plus, &buf);
+        if let (Some(region_params), true) = (&self.region_params, factor < threshold) {
```
madninja, line 185: Pull the region param and early return out sooner? or perhaps turn this into a function like 

`async fn derive_secondary_beacon(&self..) -> Result<Option<Beacon>>`

Which would give you a single error to match on, log and return instead of the many matches in this caller.. Then Option<Beacon> gets you a beacon or not if the threshold wasn't met

Vagabond:
I've reworked this to use xxhash and avoid bignums

beacon/src/entropy.rs
```diff
@@ -36,6 +36,18 @@ impl Entropy {
         })
     }
 
+    pub fn from_data(data: Vec<u8>) -> Result<Self> {
```
jeffgrunewald, line 39: is there any way for this to be (or _should_ this be) size bound in the way that the `local()` entropy Vec is bound by the size of the `LOCAL_ENTROPY_SIZE` const?

JayKickliter: Whether or not it makes sense to be bounded, it does make sense to be a `Vec` since this fn is taking ownership.

beacon/src/entropy.rs
```diff
@@ -36,6 +36,18 @@ impl Entropy {
...
+        let timestamp = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map_err(Error::from)?
+            .as_secs() as i64;
+        Ok(Self {
+            version: 1, // marked as local
```
jeffgrunewald, line 45: what does "marked as local" mean if it's a different version of entropy than the actual local entropy i.e. `version: 0`? is this to differentiate a secondary beacon's entropy from a primary beacon's entropy but it's still traceable back to 4 bytes from the primary beaconer's OS? are these distinctions captured anywhere other than in the comment above such as the `entropy.proto`?

Vagabond: yes, I wanted to distinguish the entropy source from entropy from the entropy server, I did not make a comment upstream in the proto repo yet.

madninja (approved):

PR #323, opened 2022-10-19
return tx power and tmst to poc reports

jeffgrunewald:
Returns tmst and tx_power/powe fields from transmitted beacons to the beaconer caller to update poc beacon reports with the actual values used in the transmission, as well as pulling the tmst value from poc witness packets into the associated witness reports. This should enable creation of second-order beacons

### DIFF
--- a/beacon/src/beacon.rs
+++ b/beacon/src/beacon.rs
@@ -92,6 +92,7 @@ impl TryFrom<Beacon> for poc_lora::LoraBeaconReportReqV1 {
             frequency: v.frequency,
             channel: 0,
             datarate: v.datarate as i32,
+            tmst: 0,
             tx_power: 27,
             // The timestamp of the beacon is the timestamp of creation of the
             // report (in nanos)
--- a/src/beaconer.rs
+++ b/src/beaconer.rs
@@ -5,7 +5,7 @@
 //! TODO: fuzz beacon interval to prevent thundering herd.
 
 use crate::{
-    gateway,
+    gateway::{self, BeaconResp},
     service::{entropy::EntropyService, poc::PocLoraService},
     settings::Settings,
     sync, Base64, Error, Keypair, MsgSign, Packet, RegionParams, Result,
@@ -113,15 +113,22 @@ impl Beaconer {
         };
         let beacon_id = beacon.beacon_id();
         info!(logger, "transmitting beacon"; "beacon" => &beacon_id);
-        let report = match poc_lora::LoraBeaconReportReqV1::try_from(beacon.clone()) {
+        let mut report = match poc_lora::LoraBeaconReportReqV1::try_from(beacon.clone()) {
             Ok(report) => report,
             Err(err) => {
                 warn!(logger, "failed to construct beacon report {err:?}");
                 return;
             }
         };
-        self.transmit.transmit_beacon(beacon).await;
-
+        let (powe, tmst) = match self.transmit.transmit_beacon(beacon).await {
+            Ok(BeaconResp { powe, tmst }) => (powe, tmst),
+            Err(err) => {
+                warn!(logger, "failed to transmit beacon {err:?}");
+                return;
+            }
+        };
+        report.tx_power = powe;
+        report.tmst = tmst;
         let _ = PocLoraService::new(self.poc_ingest_uri.clone())
             .submit_beacon(report, self.keypair.clone())
             .inspect_err(|err| info!(logger, "failed to submit poc beacon report: {err:?}"; "beacon" => &beacon_id))
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -1,4 +1,7 @@
-use crate::{beaconer, router::dispatcher, Error, Packet, RegionParams, Result, Settings};
+use crate::{
+    beaconer, error::RegionError, router::dispatcher, sync, Error, Packet, RegionParams, Result,
+    Settings,
+};
 use beacon::Beacon;
 use futures::TryFutureExt;
 use lorawan::PHYPayload;
@@ -18,10 +21,24 @@ pub const DOWNLINK_TIMEOUT_SECS: u64 = 5;
 pub const UPLINK_TIMEOUT_SECS: u64 = 6;
 
 #[derive(Debug)]
+pub struct BeaconResp {
+    pub powe: i32,
+    pub tmst: u32,
+}
+
+#[derive(Debug)]
 pub enum Message {
     Downlink(Packet),
-    TransmitBeacon(Beacon),
+    TransmitBeacon(Beacon, sync::ResponseSender<Result<BeaconResp>>),
     RegionParamsChanged(RegionParams),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum GatewayError {
+    #[error("unknown beacon tx power")]
+    NoBeaconTxPower,
+    #[error("beacon transmit failed")]
+    BeaconTxFailure,
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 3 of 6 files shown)


beacon/src/beacon.rs
```diff
@@ -92,6 +92,7 @@ impl TryFrom<Beacon> for poc_lora::LoraBeaconReportReqV1 {
             frequency: v.frequency,
             channel: 0,
             datarate: v.datarate as i32,
+            tmst: 0,
```
jeffgrunewald, line 95: this felt like as good a default value as any since we override the values immediately after the report is created but before sending to the ingestor

src/beaconer.rs
```diff
@@ -113,15 +125,33 @@ impl Beaconer {
...
+                "Beacon transmitted with power {powe:?} and time {tmst:?}"
+            );
+            (powe, tmst)
+        } else {
+            warn!(logger, "failed to transmit beacon");
+            return;
+        };
 
+        report.tx_power = powe.unwrap_or(27);
+        report.tmst = tmst.unwrap_or(0);
```
jeffgrunewald, line 154: If i understand correctly these should always be supplied as `Some(integer)` in the context of the beacon packets but I'd like to have a better fallback value here that doesn't feel arbitrary.

src/packet.rs
```diff
@@ -199,11 +199,16 @@ impl Packet {
                 )));
             }
         };
+        let tmst = if let Some(rx2) = &self.rx2_window {
```
jeffgrunewald, line 202: i'm not sure if this precedence logic is correct since the `to_pull_resp` method above takes an additional argument to direct it whether or not to use the `rx2_window` field if present.

src/gateway.rs
```diff
@@ -189,17 +193,20 @@ impl Gateway {
         }
     }
 
-    async fn handle_transmit_beacon(&mut self, logger: &Logger, beacon: Beacon) {
+    async fn handle_transmit_beacon(&mut self, logger: &Logger, beacon_req: BeaconReq) {
+        let BeaconReq { beacon, response } = beacon_req;
         let tx_power = if let Some(tx_power) = self.tx_power(logger) {
             tx_power
         } else {
+            response.send(Err(Error::custom("no tx power")), logger);
```
jeffgrunewald, line 201: these two custom errors are "unusual" but i'd still like to provide a better error response here, however i'm not completely sure where to create the module-level error for these: whether they belong in the `gateway.rs` module b/c that's where the error is being generated or in the `beaconer.rs` module b/c that's where they're handled and relevant.

madninja: There's a RegionError in error.rs. This has a single enum entry for NoRegionParams. Add another entry there for `NoRegionTxPower`

I also believe we need to adjust `tx_power()` to return these RegionErrors and have the two callers of them log and fire up the error in the beacon case.

src/beaconer.rs
```diff
@@ -25,6 +25,18 @@ pub enum Message {
...
 
+#[derive(Debug)]
+pub struct BeaconReq {
+    pub beacon: beacon::Beacon,
+    pub response: sync::ResponseSender<Result<BeaconResp>>,
+}
+
+#[derive(Debug)]
+pub struct BeaconResp {
+    pub powe: Option<i32>,
```
madninja, line 36: Afaict these are never set to None in the responder, so why declare them as Option here?

src/packet.rs
madninja: Yeah that's not right. the rx2 window is only used if rx1 fails. 

1. Have this function take a timestamp to declare the report at. Then have the code handling the BeaconResp use the timestamp it gets back
2. Make the timestamp use now to get the current time for the witness report

I think the first option seems closest to right?

beacon/src/beacon.rs
jeffgrunewald: i think you could say the same for the `tx_power` being set to 27 here as well (i believe that's supposed to be the max supported?); do we want to leave that as-is or use the same mechanism to default it there too with the understanding that we intend to override in any circumstance?

src/packet.rs
jeffgrunewald: @Vagabond correct me if i misunderstood but i thought this field was intended to track the "radio clock" which wouldn't be the same thing as the system clock and i'm guessing rust would be deriving `now()` from the system clock? so where should the code handling the beacon resp get such a timestamp? wouldn't it get the timestamp from the beacon which would be what i have it doing right now and just switch the precedence order so it uses `self.timestamp` first and if that's `None` use the rx2 version?

src/beaconer.rs
```diff
@@ -25,6 +25,18 @@ pub enum Message {
     RegionParamsChanged(RegionParams),
 }
 
+#[derive(Debug)]
+pub struct BeaconReq {
```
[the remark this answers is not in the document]
jeffgrunewald: which beacon message do you mean? the beacon itself is defined in the `beacon.rs` module but the enum `Message` in `gateway.rs` had a `TransmitBeacon(Beacon)` that i converted to `TransmitBeacon(BeaconReq)` if that's what you mean

src/beaconer.rs
```diff
@@ -113,15 +125,33 @@ impl Beaconer {
...
-        let report = match poc_lora::LoraBeaconReportReqV1::try_from(beacon.clone()) {
+        let mut report = match poc_lora::LoraBeaconReportReqV1::try_from(beacon.clone()) {
             Ok(report) => report,
             Err(err) => {
                 warn!(logger, "failed to construct beacon report {err:?}");
                 return;
             }
         };
-        self.transmit.transmit_beacon(beacon).await;
+        let (tx, rx) = sync::response_channel();
```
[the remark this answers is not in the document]
jeffgrunewald: or do you mean off the `gateway::MessageSender` since that's the struct the beaconer has saved for sending messages _to_ the gateway?

src/packet.rs
Vagabond: tmst is used a few different ways, rx2 only matters on scheduling a transmit, but in this case we've attached the time of the *successful* transmit to the txpk_ack, so I don't understand what were doing with rx2 here

beacon/src/beacon.rs
madninja: True.. leave as 0 then. works fine

src/gateway.rs
madninja: hmm.. ok we could create a BeaconError enum to capture just the errors related to beacon _BUT_ have the region params error still be its own thing since `tx_power()` is used for both beacons and normal downlinks?

src/packet.rs
madninja: Resolve by removing for now and just using self.timestamp

jeffgrunewald:
all (currently) outstanding comments resolved in [4c90fd7](https://github.com/helium/gateway-rs/pull/323/commits/4c90fd76c294491a0d806eeb0adf218ad5c13b5f) @madninja

src/packet.rs
```diff
@@ -202,8 +202,8 @@ impl Packet {
         let report = poc_lora::LoraWitnessReportReqV1 {
             pub_key: vec![],
             data: payload,
+            tmst: self.timestamp as u32,
             timestamp: self.timestamp,
```
Vagabond, line 206: self timestamp is 2 fields here? `tmst` should be the `tmst` from the rxpk packet and `timestamp` should be the system timestamp in nanoseconds?

madninja (commented):
Looks fine except for @Vagabond 's comment

src/packet.rs
lthiery: yea I think the `Packet` needs to have the fields added and then the `impl From<RxPk> for Packet` will need to use rxpk.get_timestamp() to get tmst and rxpk.get_time() to get UTC string. Apologies again for the crappy function names.

jeffgrunewald: Can you confirm we’re talking about a `push_data::RxPk` here? In which case the timestamp field we’re creating _is_ the tmst if I read this line correctly: https://github.com/helium/gateway-rs/blob/0fae27b7939e2051e02b56eaffef09ac4dd3582d/src/packet.rs#L53

in which case I need to add the tmst field to the proto and then re-assign this `*rxpk.get_timestamp` to that field and use the other time getter for the existing field, right?

lthiery: Yes, your understanding is correct, however, you'll want to check normal uplinks bc they might have logic hanging off of Packet.timestamp that assumes it's the 32-bit counter (which impacts downlink scheduling).

lthiery: One idea is to leave `Packet.timestamp` as is and to create `Packet.datetime` which will take the UTC string which is the new data.

jeffgrunewald: That’s a good suggestion; will do

jeffgrunewald: comment added to get the proper datetime value added to packet.proto and then get the necessary value from rxpk and convert accordingly as a future todo; temporary solution to set the field only on the witness report from the receiving gateway's systemtime::now in place

madninja (approved):

Vagabond (approved):

PR #328, opened 2022-12-08
Add some jitter to beacon intervals

JayKickliter:
This PR closes off a long-standing TODO in the beacon module. The motivation is to reduce the possibility of a thundering herd situation where gateways all beacon at the same time. It shouldn't be possible generally, because the previous beacon timer was relative to start-up time, but this additional jitter reduces the risk.

Requested by @Vagabond

### DIFF
--- a/src/gateway.rs
+++ b/src/gateway.rs
@@ -106,7 +106,7 @@ impl Gateway {
             messages,
             beacon_handler,
             listen_address: settings.listen.clone(),
-            udp_runtime: UdpRuntime::new(&settings.listen).await?,
+            udp_runtime: UdpRuntime::new(&settings.listen).await.map_err(Box::new)?,
             region_params: None,
         };
         Ok(gateway)
--- a/lorawan/src/error.rs
+++ b/lorawan/src/error.rs
@@ -11,10 +11,10 @@ pub enum LoraWanError {
 impl fmt::Display for LoraWanError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            LoraWanError::InvalidPacketType(v) => write!(f, "Invalid packet type: {:#02x}", v),
+            LoraWanError::InvalidPacketType(v) => write!(f, "Invalid packet type: {v:#02x}"),
             LoraWanError::InvalidFPortForFopts => write!(f, "Invalid: fport 0 with fopts"),
             LoraWanError::InvalidPacketSize(mtype, s) => {
-                write!(f, "Invalid packet size {} for type {:?}", s, mtype)
+                write!(f, "Invalid packet size {s} for type {mtype:?}")
             }
             LoraWanError::Io(err) => err.fmt(f),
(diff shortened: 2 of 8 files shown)


jeffgrunewald (approved):
looks like you're getting dinged by a recent clippy change given the offending file has been modified?

JayKickliter:
> looks like you're getting dinged by a recent clippy change given the offending file has been modified?

Yep. I'm going to add those fixes to this branch

madninja (commented):
Why are you jittering on every tick instead of just adding some random delay to the first tick? There's no real need to jitter on each tick at all right?

JayKickliter:
> Why are you jittering on every tick instead of just adding some random delay to the first tick? There's no real need to jitter on each tick at all right?

there's no _need_ to, but jittering every tick will make the fleet of hotspots have a very wide distribution. Or to put it another way: no clumps. We could achieve the same thing with a single delay as you suggest, but that delay would need to have a very large upper limit in order to get a good in-time spread. A very large initial delay could cause a large percentage of hotspots to not beacon if they get into some sort of boot loop through no fault of their own.

I think a better question is "does continuous randomness make things worse?" If yes, then let's have only a single random delay.

madninja:
> > Why are you jittering on every tick instead of just adding some random delay to the first tick? There's no real need to jitter on each tick at all right?
> 
> there's no _need_ to, but jittering every tick will make the fleet of hotspots have a very wide distribution. Or to put it another way: no clumps. We could achieve the same thing with a single delay as you suggest, but that delay would need to have a very large upper limit in order to get a good in-time spread. A very large initial delay could cause a large percentage of hotspots to not beacon if they get into some sort of boot loop through no fault of their own.
> 
> I think a better question is "does continuous randomness make things worse?" If yes, then let's have only a single random delay.

1. a continuous boot loop doesn't mean they'll beacon at all either with this approach
2. does it make it works? It _could_ if the jitter interval in the code is not aligned with what the verifier expects which is a relationship that's going to get lost easily

JayKickliter:
I reverted to a much simpler 'constant' radomized jitter value. AKA: we compute the random jitter once, and continuously use that value

src/gateway.rs
```diff
@@ -106,7 +106,7 @@ impl Gateway {
             messages,
             beacon_handler,
             listen_address: settings.listen.clone(),
-            udp_runtime: UdpRuntime::new(&settings.listen).await?,
+            udp_runtime: UdpRuntime::new(&settings.listen).await.map_err(Box::new)?,
```
madninja, line 109: do you actually need this? the `from` in the error definition doesn't do this for you already?

JayKickliter: Yes. Errors otherwise.

The from can only convert from the exact type. If it was used in multiple places I'd manually implement From to do the boxing, but this is the only spot

jeffgrunewald (approved):

madninja (approved):

PR #334, opened 2022-12-16
Pin b64 to v0.13

Oliv4945:
base64 crate has been updated to [v0.20](https://github.com/marshallpierce/rust-base64/releases/tag/v0.20.0) last week, breaking the API compatibility. As this new API is [still questioned](https://github.com/marshallpierce/rust-base64/issues/205), and migration [not yet documented](https://github.com/marshallpierce/rust-base64/issues/206#issuecomment-1354047328) I figure that it is safer to pin the version to `0.13` instead of `0`


madninja:
Given that this crate is still in alpha, I prefer to actually try to move the code forward instead of pinning in the past.

PR #342, opened 2023-01-16
smartharvest target: updating order in start sequence

greenbreakfast:
Our LoRa Packet Forwarder program now relies on gateway-rs to provide the regulatory region for the radio. This change makes gateway-rs start before the lora pkt fwd program.


greenbreakfast:
Got it, thanks. Will move the smart harvest specific files to our own repo. 
Closing this PR.

PR #344, opened 2023-01-19
Simplify to use settings.toml

madninja:
This simplifies helium_gateway to use a single settings.toml file. 

**NOTE** This is a breaking CLI change if you're using the `-c` flag to pass in where the configuration file is stored. The `-c` flag now needs to be a path to the actual settings.toml file


config/settings.toml
```diff
@@ -0,0 +1,53 @@
+## gateway-rs configuration file
+
+## Keypair is a string that supports specifying different locations for 
+## the secrets. The default one is file based one since all devices 
+## can support it.
+##
+## Override the setting by putting a keypair line in the
```
JayKickliter, line 7: no longer accurate?

PR #345, opened 2023-01-20
Use a reconnect backoff to back of on errors

madninja:
Instead of waitint 30 minutes on a failure to (re)connect to the packet router, use an exponential backoff on errors.

### DIFF
--- a/src/router.rs
+++ b/src/router.rs
@@ -2,17 +2,24 @@ use crate::{
     error::RegionError, gateway, region_watcher, service::router::RouterService, sync, Base64,
     Error, Keypair, MsgSign, Packet, Region, RegionParams, Result, Settings,
 };
+use exponential_backoff::Backoff;
 use helium_proto::services::router::{PacketRouterPacketDownV1, PacketRouterPacketUpV1};
 use slog::{debug, info, o, warn, Logger};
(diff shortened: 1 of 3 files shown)


src/router.rs
```diff
@@ -2,17 +2,24 @@ use crate::{
...
-use tokio::time::{self, Duration};
+use std::{collections::VecDeque, sync::Arc, time::Instant as StdInstant};
+use tokio::time::{self, Duration, Instant};
 
-pub const STORE_GC_INTERVAL: Duration = Duration::from_secs(60);
-const RECONNECT_INTERVAL: Duration = Duration::from_secs(1800); // 30 minutes
+const STORE_GC_INTERVAL: Duration = Duration::from_secs(60);
+
+const RECONNECT_BACKOFF_RETRIES: u32 = 20;
+const RECONNECT_BACKOFF_MIN_WAIT: Duration = Duration::from_secs(5);
```
jeffgrunewald, line 14: Should this min back off const be public and the reference in the region watcher refer to it or vice versa or should they stay independent of each other?

jeffgrunewald (approved):

src/router.rs
madninja: They're independent

PR #352, opened 2023-02-02
Support legacy beacons

madninja:
This replaces beacon construction with the “legacy” beacon construction used by the erlang miner.


beacon/src/beacon.rs
```diff
@@ -1,38 +1,38 @@
...
     /// and local entropy are checked for version equality.
     ///
-    /// Version 0 beacons use a Sha256 of the remote and local entropy (data and
-    /// timestamp), which is then used as a 32 byte seed to a ChaCha12 rng. This
-    /// rng is used to choose a random frequency from the given region
-    /// parameters and a payload size between MIN_BEACON_V0_PAYLOAD_SIZE and
-    /// MAX_BEACON_V0_PAYLOAD_SIZE.
+    /// Version 0/1 beacons use a Sha256 of the remote and local entropy (data
+    /// and timestamp), which is then used (truncated) as the beacon payload.
+    /// The freuency is derived from the first two bytes of the beacon payload,
```
ke6jjj, line 29: s/freuency/frequency

src/service/gateway.rs
```diff
@@ -65,10 +65,10 @@ impl Response for GatewayRespV1 {
     fn region_params(&self) -> Result<RegionParams> {
         match &self.msg {
             Some(gateway_resp_v1::Msg::RegionParamsStreamedResp(params)) => {
-                RegionParams::try_from(params.to_owned())
+                Ok(RegionParams::try_from(params.to_owned())?)
```
jeffgrunewald, line 68: surprised clippy doesn't have a problem with this; does the unwrap contain and convert the error case to the context of this module so it's a different error when it reaches the caller that if you just return the raw result of the `try_from`?

beacon/src/beacon.rs
```diff
@@ -42,24 +42,30 @@ impl Beacon {
...
+                data.truncate(BEACON_PAYLOAD_SIZE);
+
+                // Selet frequency based on the the first two bytes of the
+                // beacon data
+                let freq_seed = LittleEndian::read_u16(&data) as usize;
+                let frequency =
+                    region_params.params[freq_seed % region_params.params.len()].channel_frequency;
+
+                let spreading = region_params.select_spreading(data.len())?.as_str_name();
+                let bandwidth = region_params.bandwitdh()? / 1000; // in khz
```
ke6jjj, line 56: Does this really compile? The member is named `bandwitdh`? Yiesk.

beacon/src/beacon.rs
```diff
@@ -42,24 +42,30 @@ impl Beacon {
...
+
+                let datarate = DataRate::from_str(&format!("{spreading}BW{bandwidth}"))
+                    .map_err(|_| Error::no_data_rate())?;
+                let conducted_power = region_params.max_conducted_power()?;
```
ke6jjj, line 60: Are we storing max **conducted power** in the region or are we storing max **effective isotropic radiator power**? I'm worried this is getting mixed again.

jeffgrunewald (approved):

beacon/src/region.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.params
+            .iter()
+            .max_by_key(|p| p.max_eirp)
+            .ok_or_else(Error::no_region_params)
+            .map(|v| Decimal::new(v.max_eirp as i64, 1))
+    }
+
+    pub fn max_conducted_power(&self) -> Result<u32> {
+        self.max_eirp().and_then(|max_eirp| {
+            (max_eirp - self.gain)
```
ke6jjj, line 140: Seeing this helps me feel more relieved about my earlier comment, but what is self.gain in this function? Is that the local gateway's asserted antenna gain?

beacon/src/beacon.rs
andymck, line 29: nit 'freuency ' typo

beacon/src/region.rs
```diff
@@ -0,0 +1,203 @@
...
+            // Ensure spreading parameters are available
+            .and_then(|params| {
+                params
+                    .spreading
+                    .as_ref()
+                    .ok_or_else(Error::no_region_spreading)
+            })
+            // Find the first tagged spreading that can manage the given
+            // packet size and extract the region spreading
+            .and_then(|spreading| {
```
ke6jjj, line 171: As we guaranteed that the list of spreading factors will always be properly sorted before reaching this point? This could break if that assumption is broken.

beacon/src/region.rs
```diff
@@ -0,0 +1,203 @@
...
+                    .ok_or_else(|| Error::no_region_spreading_for_size(packet_size))
+            })
+            // Convert to RegionSpreading
+            .and_then(|region_spreading| {
+                RegionSpreading::from_i32(region_spreading)
+                    .ok_or_else(|| Error::unsupported_region_spreading(region_spreading))
+            })
+    }
+
+    pub fn bandwitdh(&self) -> Result<u32> {
```
jeffgrunewald, line 188: this is why it compiled @ke6jjj ; it's a _consistent_ typo 😁

ke6jjj: No chance to fix, eh? it's in a different crate?

beacon/src/beacon.rs
```diff
@@ -1,38 +1,38 @@
...
+pub const BEACON_PAYLOAD_SIZE: usize = 52;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Beacon {
     pub data: Vec<u8>,
 
     pub frequency: u64,
     pub datarate: DataRate,
     pub remote_entropy: Entropy,
```
andymck, line 18: as the verifier will have to reconstruct the beacon from submitted report data and then compare for equality, are the timestamps of the entropy going to cause issues ?  ie the beacon creation assigns timestamps of now() and none of those timestamps make it through to the report.

Thus when i convert from the received beacon_report to a beacon, i will likely have to assign current timestamps or use the timestamp from the report...but either way they wont match

Will PartialEq allow comparison of the timestamps to be skipped ?

src/service/gateway.rs
madninja: right.. `RegionParams::try_from` returns a `beacon::Error` which doesn't auto convert to a gateway-rs Error. 

Note though this is _not_ an unwrap..

beacon/src/beacon.rs
madninja: you just keep coming up with new names :-)

beacon/src/region.rs
madninja: yes that's the hotspots's asserted antenna gain. When gateway-rs gets "region_params" for it's public key it gets back the asserted region, gain and region_params for the region

beacon/src/region.rs
madninja: this matches the erlang code, which is what we were going for

beacon/src/beacon.rs
ke6jjj: Resolving. It is indeed conducted power being returned. The function handles the conversion correctly.

beacon/src/beacon.rs
madninja: huh? no.. the `Beacon` itself has no notion of timestamps..only the reports. You can reconstruct a beacon from a report right?

andymck: right but the beacon consists of a local and remote entropy field, each of which are of type Entropy which itself consists of a timestamp

madninja: but only the entropy _data_ is used not the timestamp

andymck: so maybe its all good then but my question is with PartialEq will those timestamps get in the way ?

ke6jjj (approved):

PR #353, opened 2023-02-03
Add multi-arch docker image compatibility.

JayKickliter:
This PR modifies the Dockerfile so that it can be used to produce single images that support multiple architectures. Currently, it supports [`linux/arm64`, `linux/amd64`]. The restriction on those two is only because I chose to perform phase 1 of the build on the host architecture, and not in QEMU.

Example invocation to an image that can run on amd64 and aarch64:

```sh
docker buildx build --platform linux/arm64,linux/amd64 .
```

I left notes in the Dockerfile on how to add additional architecture support. However, if we and our CI infrastructure are fine with the **extremely** long build times, we can modify the Dockerfile to perform both build phases in QUEMU and be compatible with any platforms that `buildx` supports.

# Testing

The only thing can't be tested until merging is if tagging a release results in a properly named tag in the miner quay. That said, publishing to both quay.io/repository/team-helium/miner and quay.io/repository/team-helium/test-images is working.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,3 +95,16 @@ jobs:
         with:
           fail_on_unmatched_files: true
           files: helium-gateway-*.tar.gz
+
+  buildx:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup Docker buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          platforms: linux/amd64,linux/arm64
+
+      - name: Build multi-architecture Docker image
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -95,3 +95,44 @@ jobs:
         with:
           fail_on_unmatched_files: true
           files: helium-gateway-*.tar.gz
+
+  docker_buildx:
```
JayKickliter, line 99: I'm thinking it may make most sense to moving the git-tag-release triggered build, wich published to quay.io/repository/team-helium/miner, to the release job. And leave the rest as is to publish every build to quay.io/repository/team-helium/test-images

PR #355, opened 2023-02-03
The seed isn’t enough to generate the data

madninja:
* This fixes up the beacon data the way it was intended.. the entropy is used to seed an RNG which then generates a random bytes of the requested length
* Sets the required length to 51 to allow for the lora header

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -208,6 +208,7 @@ dependencies = [
  "lazy_static",
  "prost",
  "rand",
+ "rand_chacha",
  "rust_decimal",
  "serde",
  "serde_json",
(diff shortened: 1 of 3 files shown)


ke6jjj (approved):
Looks better. I don't know `rand_chacha:...::from_seed()`, of course, but if it generates 51 bytes in the end then good?

madninja:
> Looks better. I don't know `rand_chacha:...::from_seed()`, of course, but if it generates 51 bytes in the end then good?

Yeah it implements the `Rng` trait which implements the `sample_iter` function.

ke6jjj (approved):

PR #360, opened 2023-02-13
use portable byte order for beacon data generation

andymck:


beacon/src/entropy.rs
```diff
@@ -43,7 +43,7 @@ impl Entropy {
 
     pub(crate) fn digest<D: Digest>(&self, state: &mut D) {
         state.update(&self.data);
-        state.update(self.timestamp.to_ne_bytes());
+        state.update(self.timestamp.to_be_bytes());
```
madninja, line 46: Use to_le

andymck: before deciding be or le, i checked elsewhere on the oracles and gateway rs and we always choose to_be.  So was sticking with that convention ?

madninja: Hmm you’re right. It needs to lock an endiannesss. Most hotspots are le though so you’d expect only a smaller fraction to fail?

andymck: yeah i am not sure this is the actual issue we are seeing atm on devnet irrespectively we should lock it one way or the other to ensure nothing arises in the future

madninja: yeah, to_ne is surely not right

PR #364, opened 2023-02-21
"validator" feature

madninja:
This adds a "validator" feature (enabled by default) which when enabled:

* Uses the lgw seeds and gateway service to select a validator to get a routing information stream from
* Uses lgw seeds and `region_for` to watch for region changes

Disabling/removing the validator feature will turn on the config service for region watching as well as switch to packet_router as the simpler/ultimate goal for packet routing. Disabling validator should happen in early March 2023


src/api/server.rs
```diff
@@ -59,11 +57,9 @@ impl Api for LocalServer {
 
     async fn region(&self, _request: Request<RegionReq>) -> ApiResult<RegionRes> {
         let region_params = self.region_watch.borrow();
-        let region = region_params
-            .as_ref()
-            .map_or_else(|| self.default_region, |params| params.region)
-            .into();
-        Ok(Response::new(RegionRes { region }))
+        Ok(Response::new(RegionRes {
```
JayKickliter, line 60: I see no error path. Why return a `Result`?

src/beaconer.rs
```diff
@@ -58,7 +61,7 @@ pub struct Beaconer {
     /// The last beacon that was transitted
     last_beacon: Option<beacon::Beacon>,
     /// Use for channel plan and FR parameters
-    region_params: Option<RegionParams>,
+    region_params: RegionParams,
```
JayKickliter, line 64: what has changed that lets us avoid wrapping params in `Option` all over now?

src/error.rs
```diff
@@ -81,6 +81,8 @@ pub enum ServiceError {
     Channel,
     #[error("no service")]
     NoService,
+    #[error("age {age}s > {max_age}s")]
```
JayKickliter, line 84: the naming of this new variant `Check` isn't obvious in context of the age-related error message.

JayKickliter: On the other hand, hyper-focused error variants can be cumbersome.

src/message_cache.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    pub fn store(&mut self, message: T, received: Instant) -> Result {
+        self.waiting.push_back(CacheMessage { message, received });
+        if self.waiting_count() > self.max_messages as usize {
+            self.waiting.pop_front();
+        }
+        Ok(())
+    }
+
+    pub fn pop_waiting(&mut self) -> Option<CacheMessage<T>> {
```
JayKickliter, line 54: Even on wrapper types like this message cache I personally prefer to keep the naming of the underlying type doing the heavy lifting. AKA: I'd prefer `MessageCache::pop_front` and `MessageCache::len`

src/message_cache.rs
```diff
@@ -0,0 +1,70 @@
...
+        self.waiting.pop_front()
+    }
+
+    pub fn waiting_count(&self) -> usize {
+        self.waiting.len()
+    }
+
+    /// Removes waiting packets older than the given duration. Returns the number
+    /// of packets that were removed.
+    pub fn gc(&mut self, duration: Duration) -> usize {
```
JayKickliter, line 64: Based on my previous comments, it's probably obvious that I don't think this function should exist as is. At a minimum, I don't think it should be public.

src/message_cache.rs
```diff
@@ -0,0 +1,70 @@
...
+            max_messages,
+        }
+    }
+
+    pub fn store(&mut self, message: T, received: Instant) -> Result {
+        self.waiting.push_back(CacheMessage { message, received });
+        if self.waiting_count() > self.max_messages as usize {
+            self.waiting.pop_front();
+        }
+        Ok(())
```
JayKickliter, line 51: I see no reason for this fn to return anything.

src/message_cache.rs
```diff
@@ -0,0 +1,70 @@
...
+            waiting,
+            max_messages,
+        }
+    }
+
+    pub fn store(&mut self, message: T, received: Instant) -> Result {
+        self.waiting.push_back(CacheMessage { message, received });
+        if self.waiting_count() > self.max_messages as usize {
+            self.waiting.pop_front();
+        }
```
JayKickliter, line 50: This assumes GC has ran since the last time this function was called. It's probably better to either:

1. inline the GC here and have this fn return number of GCs messages
2. not do any popping here if the number of GCd messages is actually important to know

src/packet_router/mod.rs
```diff
@@ -172,85 +163,27 @@ impl Router {
     }
 
     async fn send_waiting_packets(&mut self, logger: &Logger) {
-        while let Some(packet) = self.store.pop_waiting_packet() {
+        while let Some(packet) = self.store.pop_waiting() {
             match self.send_packet(logger, packet).await {
```
JayKickliter, line 167: Minor nit
```suggestion
            if let Err(err) = self.send_packet(logger, packet).await {
```

src/packet_router/mod.rs
```diff
@@ -172,85 +163,27 @@ impl Router {
...
+    pub async fn mk_uplink(&self, packet: CacheMessage<Packet>) -> Result<PacketRouterPacketUpV1> {
+        let mut uplink: PacketRouterPacketUpV1 = packet.into_inner().try_into()?;
+        uplink.region = self.region_params.region.into();
         uplink.gateway = self.keypair.public_key().into();
         uplink.signature = uplink.sign(self.keypair.clone()).await?;
         Ok(uplink)
     }
 
-    async fn send_packet(&mut self, logger: &Logger, packet: QuePacket) -> Result {
+    async fn send_packet(&mut self, logger: &Logger, packet: CacheMessage<Packet>) -> Result {
```
JayKickliter, line 182: Why take the wrapped packet? I feel like `CacheMessage` should be an internal detail to the cache.

src/region_watcher.rs
```diff
@@ -8,41 +8,55 @@ const REGION_BACKOFF_RETRIES: u32 = 10;
 const REGION_BACKOFF_MIN_WAIT: Duration = Duration::from_secs(5);
 const REGION_BACKOFF_MAX_WAIT: Duration = Duration::from_secs(3600); // 60 minutes
 
-pub type MessageSender = watch::Sender<Option<RegionParams>>;
-pub type MessageReceiver = watch::Receiver<Option<RegionParams>>;
+pub type MessageSender = watch::Sender<RegionParams>;
+pub type MessageReceiver = watch::Receiver<RegionParams>;
+
+pub fn current<T>(receiver: &watch::Receiver<T>) -> T
```
JayKickliter, line 14: `current` is such an overloaded word. A more descriptive name or a comment would help here.

JayKickliter (commented):
Since this is a big PR and you're still pushing, I'm going to split my review into at least two parts. Here's part one

src/service/gateway.rs
```diff
@@ -0,0 +1,278 @@
...
+#[derive(Debug, Serialize)]
+pub struct GatewayVersion {
+    major: u16,
+    minor: u16,
+    patch: u16,
+}
+
+impl From<u64> for GatewayVersion {
+    fn from(v: u64) -> Self {
+        let patch = (v % 10000) as u16;
```
JayKickliter, line 252: is it wishful thinking to ask for a simpler versioning scheme? Like each byte be a single field?

src/router/routing.rs
```diff
@@ -0,0 +1,82 @@
...
+    filters: Vec<EuiFilter>,
+    subnets: Vec<DevAddrFilter>,
+}
+
+impl Routing {
+    pub fn contains_uri(&self, uri: &KeyedUri) -> bool {
+        self.uris.iter().any(|keyed_uri| keyed_uri == uri)
+    }
+
+    pub fn matches_routing_info(&self, routing_info: &Option<RoutingInformation>) -> bool {
```
JayKickliter, line 20: Is this not called anywhere?

beacon/src/region.rs
```diff
@@ -145,6 +145,14 @@ impl TryFrom<GatewayRegionParamsResV1> for RegionParams {
 }
 
 impl RegionParams {
+    pub fn for_region(region: Region) -> Self {
```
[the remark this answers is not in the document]
madninja: Fixed by `impl From<Region> for RegionParams`

config/settings.toml
```diff
@@ -50,3 +50,176 @@ uri = "http://mainnet-router.helium.io:8080/"
 # Maximum number of packets to queue up for the packet router
 queue = 20
 
+# Default target routers for data packets that are not known to helium packet
+# routers. These are legacy settings and will be removed when the Helium Packet
```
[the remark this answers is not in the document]
madninja: yes.. the `gateways` and `routers` section are ignored when `validator` is turned off. This will all go away once HPR and the config service is running in a few weeks here 🤞🏻

src/beaconer.rs
madninja: That first From<Region> for RegionParams allows us to construct a "default" RegionParams with no gain and empty params but a region set. This simplifies a lot at the expense of needing to know that empty params in RegionParams means you can't beacon or downlink

src/api/server.rs
madninja: To meet the tonic  contract

src/router/filter.rs
```diff
@@ -0,0 +1,149 @@
...
+        buf.put_u64_le(*appeui);
+        let mut hasher = Xxh64::new(0);
+        hasher.update(&data);
+        let hash = hasher.digest();
+        self.0.contains(&hash)
+    }
+}
+
+const BITS_23: u64 = 8388607; // biggest unsigned number in 23 bits
+const BITS_25: u64 = 33554431; // biggest unsigned number in 25 bits
```
JayKickliter, line 55: Could derive or use binary literal instead of a decimal:
```rust
const BITS_25: u64 = 2_u64.pow(25) - 1; // biggest unsigned number in 25 bits
```
```rust
const BITS_25: u64 = 0b0001_1111_1111_1111_1111_1111_1111; // biggest unsigned number in 25 bits
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=16cfbc866b3b73e70420a1557944a7dd

src/error.rs
madninja: Yeah.. I was being lazy and figured I'd leak this error from the `validator` feature to maybe use it later. There also wasn't a good way to feature flag the error entry out without the error attribute macro blowing up

src/message_cache.rs
madninja: I like it.. renaming them to pop_front, push_back and len

src/packet_router/mod.rs
madninja: Because `hold_time` comes out of the CacheMessage

src/service/gateway.rs
madninja: Erlang miner compatibility.. only there until validators go away.

src/region_watcher.rs
madninja: Renamed to `current_value`

src/message_cache.rs
madninja: Concur.. fixed

src/packet_router/mod.rs
madninja: yup, fixed

src/router/routing.rs
madninja: Called here https://github.com/helium/gateway-rs/blob/madninja/router_support/src/router/dispatcher.rs#L292

src/message_cache.rs
madninja: Fixed in [433ce5b](https://github.com/helium/gateway-rs/pull/364/commits/433ce5bf279a402bae426365020d265f03aca024)

src/message_cache.rs
madninja: [433ce5b](https://github.com/helium/gateway-rs/pull/364/commits/433ce5bf279a402bae426365020d265f03aca024)

PR #366, opened 2023-02-22
Bump proto dep

vihu:
Prep for bumping gateway-rs in oracles repo.


vihu:
Closing as we'll bump proto in https://github.com/helium/gateway-rs/pull/364 itself

PR #373, opened 2023-02-24
Remove unused min conducted power

madninja:


JayKickliter (approved):
As long as the builds pass (proving unused) go ahead at merge at will

PR #374, opened 2023-02-28
Add TPM support to the dockerfile

isergieienkov:


madninja (commented):
Would it not be preferable to have the Dockerfile create a docker image that supports tpm and etc in the same image instead of having to generate two images, one of which is hidden?

JayKickliter:
@madninja yes, it would be. For some reason I thought those were mutually exclusive. If not, we should definitely do it because this build setup is going to get messy fast.

JayKickliter:
Closing in favor of #376

PR #376, opened 2023-03-01
Enable TPM+ECC support in docker builds

JayKickliter:
I transferred @isergieienkov branch for #374 to this repo so we can test CI.

I was defeated in my attempt to cross-compile with TPM support. So now the build phase happens under QEMU when building for architectures that differ from the host (notice the removal of `--from $BUILDPLATORM`). The downside is that this can be really slow. Additionally, it requires nightly because QEMU OOMs when updating the cargo cache unless we use sparse-registry.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,7 +61,6 @@ FROM alpine:3.17.1
 ENV RUST_BACKTRACE=1
 ENV GW_LISTEN="0.0.0.0:1680"
 RUN apk add --no-cache --update \
-    curl \
     libstdc++ \
     tpm2-tss-esys \
     tpm2-tss-fapi \
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,14 @@ COPY . .
 
 ENV CC=gcc CXX=g++ CFLAGS="-U__sun__" RUSTFLAGS="-C target-feature=-crt-static"
 
-RUN cargo build --release --no-default-features --features=ecc608,tpm
+# TMP build fail when cross compiling, so we need to use QEMU when
+# building for not-host architectures. But QUEMU builds fail in CI due
+# to OOMing on cargo registry updating. Therefore, we will need to
+# compile with nightly until cargo's sparse registry stabilizes.
+RUN rustup toolchain install nightly
+ENV CARGO_UNSTABLE_SPARSE_REGISTRY=true
+
+RUN cargo +nightly build --release --no-default-features --features=ecc608,tpm
 RUN mv target/release/helium_gateway .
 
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -47,7 +47,7 @@ ENV CC=gcc CXX=g++ CFLAGS="-U__sun__" RUSTFLAGS="-C target-feature=-crt-static"
 RUN rustup toolchain install nightly
 ENV CARGO_UNSTABLE_SPARSE_REGISTRY=true
 
-RUN cargo +nightly build --release --no-default-features --features=ecc608,tpm
+RUN cargo +nightly build --release --features=tpm
```
madninja, line 50: why not also support ecc608 in one image? or is that done by default + this features list?

madninja: I believe you said it does

madninja (approved):

Dockerfile
JayKickliter: Yeah. Default enables validator and ecc

madninja: ok, may want to take it out of draft?

Dockerfile
```diff
@@ -22,38 +22,33 @@
...
-RUN \
-case "$TARGETPLATFORM" in \
-    "linux/arm64") echo aarch64-unknown-linux-musl > rust_target.txt ;; \
-    "linux/amd64") echo x86_64-unknown-linux-musl > rust_target.txt ;; \
-    *) exit 1 ;; \
-esac
+# TMP build fail when cross compiling, so we need to use QEMU when
+# building for not-host architectures. But QUEMU builds fail in CI due
+# to OOMing on cargo registry updating. Therefore, we will need to
+# compile with nightly until cargo's sparse registry stabilizes.
```
jeffgrunewald, line 46: i believe i heard this was supposed to land in the 1.68 release

Dockerfile
```diff
@@ -65,7 +60,15 @@ RUN mv target/$(cat rust_target.txt)/release/helium_gateway .
 FROM alpine:3.17.1
 ENV RUST_BACKTRACE=1
 ENV GW_LISTEN="0.0.0.0:1680"
-RUN apk add --no-cache --update curl
+RUN apk add --no-cache --update \
+    curl \
```
jeffgrunewald, line 64: we were not able to remove the dependency on the curl implementation to retrieve remote entropy from the entropy service yet or is this needed for something else?

madninja: Good catch

@JayKickliter curl is no longer required

Dockerfile
JayKickliter: Cool. Let’s try to remember to revisit this when it does

Dockerfile
JayKickliter, line 64:
```suggestion
```

PR #385, opened 2023-03-09
Tidy up CI

madninja:
* Fix release artifacts missing for cross builds
* Clean up build name steps

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,13 @@ on:
     branches: [main]
     tags: ["v*"]
 
+env:
+  CARGO_INCREMENTAL: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -4,39 +4,47 @@ on:
     branches: [main]
     tags: ["v*"]
 
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_NET_RETRY: 10
+  RUST_BACKTRACE: short
```
JayKickliter, line 10: I'm pro-always-enable-backtraces, but were we getting backtraces on CI builds?

madninja: You can get them during tests

PR #397, opened 2023-03-17
Add router as a new info key

madninja:
`helium_gateway info router` returns the router URI and whether the process is connected to the packet router


JayKickliter (commented):
I've confirmed that this new command works as expected when gwrs _is connected_ to router:

```json
{
  "router": {
    "connected": true,
    "uri": "http://mainnet-router.helium.io:8080/"
  }
}
```

and when it _is not_ connected (by giving it a bad URI)

```json
{
  "router": {
    "connected": false,
    "uri": "http://mainnet-router.helium.com:8080/"
  }
}
```

JayKickliter:
Next, I'm going to test that gateway_config can easily use this API call

PR #407, opened 2023-04-01
Update GW_LISTEN

openxzx:
GW_LISTEN is 127.0.0.1:1680 in the official Helium documentation, but GW_LISTEN is 0.0.0.0:1680 in Dockerfile, refer to the link: https://docs.helium.com/solana/migration/maker-hotspot-software/#quick-reference-summary

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,7 +59,7 @@ RUN mv target/release/helium_gateway .
 # ------------------------------------------------------------------------------
 FROM alpine:3.17.1
 ENV RUST_BACKTRACE=1
-ENV GW_LISTEN="0.0.0.0:1680"
+ENV GW_LISTEN="127.0.0.1:1680"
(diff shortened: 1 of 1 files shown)


shawaj:
You can already override this in the settings.toml...https://github.com/helium/gateway-rs/blob/d862bc9962866e575d7c6dd731d65cf849a586ed/config/settings.toml#L15

JayKickliter:
> So why does it need an env variable in the dockerfile at all?

It technically doesn't, but changing it at this point will likely be a breaking change for some makers.

JayKickliter:
> So why does it need an env variable in the dockerfile at all?

It technically doesn't, but changing it at this point will likely be a breaking change for some makers.

madninja:
I'm going to close this as won't fix for now

openxzx:
Okay, thank you for your reply.

PR #411, opened 2023-04-06
Add support for port or complete api socketaddr

madninja:
This adds support for the local listen address to be either a port or a fully specified `ip:port` listen address. It is _RECOMMENDED_ to only use the port configuration (which defaults the ip to local host listening only) to avoid exposing local apis to the open internet.


madninja:
> Can it be a hostname as well such as helium-miner:4467 or does it need to be an IP?

It's a rust SocketAddr so pretty sure that needs to be an IP

Issue #408, opened 2023-04-04
Radio region should default to No-Region at Startup


ke6jjj:
# Summary

Gateway should default to no-region as a factory default.

# Details

1. Gateway-rs should support a "no region set" mode, wherein no region information is known but the daemon continues to run in idle mode, attempting to configure itself from the Configuration Service.
2. Gateway-rs should support accepting a region _hint_ at startup (via `GW_REGION` environment variable) so that a Maker can quick-start the gateway with the last known region it was in, to support a best-effort attempt at keeping data flowing during a Configuration Service outage.
3. The hinting mechanism should require explicit action on behalf of the Maker, so there should be no default "region" set in the reference `settings.toml`
4. The `info region` CLI verb (and any corresponding API) should support returning JSON `null` if the region is not set (or a similar API-appropriate sentinel value).

PR #413, opened 2023-04-07
Set region to unknown on startup

madninja:
This sets the current/default region  to a new “UNKNOWN” region to avoid invalid uplink packet reports, region parameters (and thus invalid downlinks) when the gateway can’t or has not received it’s asserted region from the config service yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,15 +82,17 @@ helium-gateway pre-installed, manual installation requires you to:
    Supported values are `stdio` or `syslog`. Note you may need to configure
    the `syslog` service on your device to accept the logs.
 
-6. Configure the region if required. The default region of the gateway is
-   `US915`, if your region is different you can set the right one in the
-   `settings.toml` file.
+6. Configure the region if required. The default region of the gateway is set to
+   `UKNOWN`, and fetched from the asserted location of the gateway. Setting the
+   regioon to a known region or caching the last fetched region and using the
+   `GW_REGION` environment variable on startup will allow the gateway to use the
+   correct region for uplinks immediately, while the region parameters are
+   retrieved.
 
    The supported region values are listed in the [region protobuf definition](https://github.com/helium/proto/blob/master/src/region.proto)
 
-   **NOTE**: This regional setting is only used for uplinks. Due to TX power
-   regulations, the gateway location needs to be asserted on the blockchain to
-   be able to send downlinks.
+   **NOTE**: Due to TX power regulations, the gateway location needs to be
+   asserted on the blockchain to be able to send downlinks.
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -82,15 +82,17 @@ helium-gateway pre-installed, manual installation requires you to:
    Supported values are `stdio` or `syslog`. Note you may need to configure
    the `syslog` service on your device to accept the logs.
 
-6. Configure the region if required. The default region of the gateway is
-   `US915`, if your region is different you can set the right one in the
-   `settings.toml` file.
+6. Configure the region if required. The default region of the gateway is set to
+   `UKNOWN`, and fetched from the asserted location of the gateway. Setting the
+   regioon to a known region or caching the last fetched region and using the
```
jeffgrunewald, line 87:
```suggestion
6. Configure the region. The default region of the gateway is set to
   `UKNOWN`, and fetched from the asserted location of the gateway. Setting the
   region to a known region or caching the last fetched region and using the
```
seems required now

config/settings.toml
```diff
@@ -20,8 +20,10 @@ api = 4467
 
 # The default region to use until a region is received from the Helium network.
 # This value should line up with the configured region of the semtech packet
-# forwarder. 
-region = "US915"
+# forwarder. Note: Not setting this here or with a GW_REGION env var will stop
+# packet and poc beacong until the correct asserted region is retrieved.
```
jeffgrunewald, line 24:
```suggestion
# packet and poc beaconing until the correct asserted region is retrieved. 
```

jeffgrunewald (approved):

ke6jjj (approved):

README.md
madninja: No.. the region is not required at all.. That's the whole point of not having it.. that there _is_ no good default region so jus too get the current asserted region from chain.

madninja:
@ke6jjj only caveat to #408 is that the unknown region is printed as `"UNKNOWN"` instead of `null`. E.g. 

```
{
  "region": "UNKNOWN"
}
```

That ok?

ke6jjj:
> @ke6jjj only caveat to #408 is that the unknown region is printed as `"UNKNOWN"` instead of `null`. E.g.

Well, unfortunately no. The idea is to return a proper null for the API in question. For JSON that really would be `null`. We're looking for something consistent that Makers can program to.

madninja:
> > @ke6jjj only caveat to #408 is that the unknown region is printed as `"UNKNOWN"` instead of `null`. E.g.
> 
> Well, unfortunately no. The idea is to return a proper null for the API in question. For JSON that really would be `null`. We're looking for something consistent that Makers can program to.

Fixed

PR #415, opened 2023-04-11
feat: add builds and use prebuilt action

shawaj:
- add ecc608 based x86_64 image (LongAP uses this configuration)
- add docker builds for armv6 and armv7 via qemu
- change the build and push step to use the prebuilt docker action
- add github actions cache to builder action step to speed up future builds
- fix typos and add a bit more detail to readme

~~- gateway-mfr-rs uses aarch64-unknown-linux-gnu instead of aarch64-unknown-linux-musl so this unifies the two~~
~~- adds missing armv7-unknown-linux-gnueabihf which is built in gateway-mfr-rs~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,7 @@ jobs:
           - aarch64-unknown-linux-musl
           - mips-unknown-linux-musl
           - mipsel-unknown-linux-musl
+          - x86_64-unknown-debian-gnu
           - x86_64-tpm-debian-gnu
 
     steps:
@@ -142,10 +143,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Setup | QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Setup | Docker
         uses: docker/setup-buildx-action@v2
         with:
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6
 
       - name: Setup | Quay.io
         uses: docker/login-action@v2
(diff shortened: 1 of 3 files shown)


madninja:
They "got lost" because no-one was actually using them. Do you have a use for them or are you trying to be complete because of the README?

madninja:
> @madninja yeah we only need the additional x86 one. I was trying to be helpful but have clearly done the opposite :-P
> 
> I assumed they were necessary / used as they were previously set up as targets for actual users in the makefile (and readme): https://github.com/helium/gateway-rs/blob/2d44b5c11a6b6d80c7e5fe67ef1f94f389c7eeba/Makefile.toml
> 
> Is that not the case any more?

The affected makers are either not using the builds here or are no longer in business. If you want to fix them up you can but you'd have to get an updated protoc in the build images.

shawaj:
ok - will keep it simple and can those two for now then 👍 

thanks, as always, @madninja

have fixed and rebased

.github/workflows/ci.yml
```diff
@@ -57,6 +57,7 @@ jobs:
           - aarch64-unknown-linux-musl
           - mips-unknown-linux-musl
           - mipsel-unknown-linux-musl
+          - x86_64-unknown-linux-gnu
```
madninja, line 60: Rename to `x86_64-debian-gnu` ?

Makefile.toml
```diff
@@ -4,6 +4,9 @@ FEATURES = "ecc608"
 BUILD_COMMAND = "cross"
 TAR = { source = "${CARGO_MAKE_RUST_TARGET_OS}", default_value = "linux", mapping = { macos = "gtar", linux = "tar" } }
 
+[env.x86_64-unknown-linux-gnu]
```
madninja, line 7: name

.github/workflows/ci.yml
shawaj: fixed in https://github.com/helium/gateway-rs/pull/415/commits/6ec9e0b9a1cc0b28121ed75ac9c7e5881ef9522b

shawaj:
I guess the build is failing because i don't have access to the repository secrets but otherwise seems to be ok

.github/workflows/ci.yml
```diff
@@ -142,10 +143,13 @@ jobs:
...
           fetch-depth: 0
 
+      - name: Setup | QEMU
+        uses: docker/setup-qemu-action@v2
+
       - name: Setup | Docker
         uses: docker/setup-buildx-action@v2
         with:
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6
```
madninja, line 152: Do you actually have a need for these architectures?

shawaj: they are for raspberry pi 1/2/3 with non 64bit archs - mainly for DIY gateways but we have some hobbyists using some of our codebase for this purpose for their own DIY gateways

we are also using armv6 (pi zero) in our light hotspot

madninja (approved):

madninja:
This is an issue with your build system Ge a newer version of the protocol tool that supports proto3 format.

NkkVr:
Error: Custom { kind: Other, error: "protoc failed: blockchain_txn.proto:1:10: Unrecognized syntax identifier \"proto3\".  This parser only recognizes \"proto2\".\n" }
root@Helium:~/Helium/gateway-rs# brew upgrade protobuf

Command 'brew' not found, did you mean:

  command 'brec' from deb bplay (0.991-10build1)
  command 'qbrew' from deb qbrew (0.4.1-8build1)

Try: apt install <deb name>

root@Helium:~/Helium/gateway-rs# protoc --version
libprotoc 3.6.1
root@Helium:~/Helium/gateway-rs#

I'm trying to keep it updated

NkkVr:
I've been stopping by since night. There is no way I haven't tried. I failed. Can you give me a clue?

NkkVr:
So how can I solve this problem?

PR #417, opened 2023-04-11
CI fix

madninja:
Fixes CI after docker related changes  in #415


shawaj:
@madninja apologies, I derped on this as I misread the builder as using alpine3.17 image, which has armv6 and armv7 variants.

But the rust:alpine3.17 doesn't https://hub.docker.com/layers/library/rust/alpine3.17/images/sha256-ac075f925a687642fe4281ab97b5c017b69c84d62e79345a06dc5371628c7481?context=explore

Shall I submit a new PR backing out this change and just leaving the rest?

I'll fix this separately and then submit a new, working PR.

Apologies again

shawaj:
I think #419 should fix this

PR #419, opened 2023-04-11
fix: backout armv6/ armv7 builds

shawaj:
Fixes the issue in #415


madninja:
@shawaj Sigh.. we gotta be more careful just yolo'ing actions https://github.com/helium/gateway-rs/actions/runs/4695602550/jobs/8324943218#step:9:131

PR #423, opened 2023-04-14
fix: tag ref incompatible with GH actions

shawaj:
this should fix the tag error in https://github.com/helium/gateway-rs/actions/runs/4695602550/jobs/8324943218

```
Error: buildx failed with: ERROR: invalid tag "quay.io/team-helium/miner:gateway-${GITHUB_REF#refs/*/}": invalid reference format
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,8 +173,8 @@ jobs:
         with:
           push: true
           platforms: linux/amd64,linux/arm64
-          labels: "VERSION=${{ env.image_tag }}"
-          tags: "quay.io/team-helium/test-images:gateway-${{ env.image_tag }}"
+          labels: VERSION=${{ env.image_tag }}
+          tags: quay.io/team-helium/test-images:gateway-${{ env.image_tag }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
@@ -189,9 +189,9 @@ jobs:
         with:
           push: true
           platforms: linux/amd64,linux/arm64
-          labels: "VERSION=${GITHUB_REF#refs/*/}"
+          labels: VERSION=${{ github.ref_name }}
           tags: |
-            quay.io/team-helium/miner:gateway-${GITHUB_REF#refs/*/}
+            quay.io/team-helium/miner:gateway-${{ github.ref_name }}
             quay.io/team-helium/miner:gateway-latest
           cache-from: type=gha
           cache-to: type=gha,mode=max


shawaj:
apologies @madninja I didnt realise `${GITHUB_REF#refs/*/}` wouldn't work in the action steps context

.github/workflows/ci.yml
```diff
@@ -189,9 +189,9 @@ jobs:
         with:
           push: true
           platforms: linux/amd64,linux/arm64
-          labels: "VERSION=${GITHUB_REF#refs/*/}"
+          labels: "VERSION=${{  github.ref_name }}"
           tags: |
-            quay.io/team-helium/miner:gateway-${GITHUB_REF#refs/*/}
+            quay.io/team-helium/miner:gateway-${{  github.ref_name }}
```
JayKickliter, line 194: Either this needs quotes, or the quotes on line 177 are unnecessary. Do you have any idea?

JayKickliter: But we know the [test-images](https://quay.io/repository/team-helium/test-images?tab=tags&tag=latest) build worked, so maybe add quotes regardless?

JayKickliter, line 194: Additionally, there's a unequal number of white spaces between `{{` (2) `github.ref_name` (1) `}}`

shawaj: i think with the list setup it doesnt need quote:
```
tags: |
```

but in the other setting it does. 

let me see if i can find any docs to confirm that...

shawaj: @JayKickliter looking at these links:

https://github.com/docker/build-push-action
https://docs.github.com/en/actions/learn-github-actions/contexts 

it looks like quotes aren't needed in either setting

JayKickliter: How about temporarily moving 
```
          tags: |
            quay.io/team-helium/miner:gateway-${{  github.ref_name }}
            quay.io/team-helium/miner:gateway-latest
```
to the test-images section above to test? You can push and see if the two tags go to the test-images repo

JayKickliter: How about temporarily moving 
```
          tags: |
            quay.io/team-helium/miner:gateway-${{  github.ref_name }}
            quay.io/team-helium/miner:gateway-latest
```
to the test-images section above to test? You can push and see if the two tags go to the test-images repo

shawaj: fixed - good catch!

shawaj: actually @JayKickliter I don't think that will work if i do it as I don't have write access on the repo I think the repository secrets aren't available. As the quay.io pushes to test-images always fail with:

```
Run docker/login-action@v2
Error: Username and password required
```

shawaj: @JayKickliter I just tried it here and it worked fine without quotes:
https://github.com/NebraLtd/hm-diag/pull/547

JayKickliter: Thanks

shawaj:
> pushed into this repo as #424 to ensure the test image side works.. I was convinced the quotes were needed there

i always get confused with yaml syntax and quotes 🤣

madninja (approved):

madninja:
Here's hoping..

PR #428, opened 2023-04-28
Add settings poc enable flag

### LINKED ISSUE #427: Data Only peforming POC.
it has been noted that data-only hotspots are participating in POC (actively sending beacons, and listening and reporting witnesses),
i see some merit to keeping the witness receipts getting recorded but would it not be a idea to have beacons disabled on data-only hotspots maybe a var in gateway rs to activate beacons

madninja:
This allows poc activity to be disabled to avoid unneeded RF
traffic. Useful for dataonly hotspots.

Fixes #427

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -79,6 +79,9 @@ impl tracing_subscriber::fmt::time::FormatTime for TimeFormatter {
 /// Settings for proof-of-coverage (PoC).
 #[derive(Debug, Deserialize, Clone)]
 pub struct PocSettings {
+    // Enable/disable poc related activities (baecon/witness)
+    #[serde(default = "default_poc_enable")]
+    pub enable: bool,
     /// Entropy URL.
     #[serde(with = "http_serde::uri")]
     pub entropy_uri: Uri,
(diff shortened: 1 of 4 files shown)


src/settings.rs
```diff
@@ -79,6 +79,9 @@ impl tracing_subscriber::fmt::time::FormatTime for TimeFormatter {
 /// Settings for proof-of-coverage (PoC).
 #[derive(Debug, Deserialize, Clone)]
 pub struct PocSettings {
+    // Enable/disable poc related activities (baecon/witness)
+    #[serde(default = "default_poc_enable")]
+    pub enable: bool,
```
lthiery, line 84: why not rename this to disable? you would avoid this default override plus the only if block that uses this boolean (`handle_received_beacon`) inverts is anyway.

lthiery (approved):

lthiery:
Isn't there some logic around reporting beacons that could be disabled too?

src/settings.rs
lthiery: Oh nevermind, it's used elsewhere too. Not a big deal either way

lthiery (approved):

PR #430, opened 2023-05-23
uplink implement Send

michaeldjeffrey:


src/settings.rs
```diff
@@ -237,7 +237,7 @@ pub mod log_level {
     use std::fmt;
 
     #[derive(Debug, Clone, Copy)]
-    pub struct Level(tracing::Level);
+    pub struct Level(pub tracing::Level);
```
madninja, line 240: Why do you need this? I’d use a From trait or Deref one to get here. But since you’re not using what you changed?

michaeldjeffrey: This allows constructing `Settings` without needing to go through the full file parsing.

madninja (approved):

PR #437, opened 2023-07-21
Filter out non parseable lorawan packets

madninja:
This filters out packets received from the packet forwarder that are not valid lorawan packets

### DIFF
--- a/src/packet.rs
+++ b/src/packet.rs
@@ -121,7 +121,11 @@ impl PacketUp {
         // An uplinkable packet is a parseable lorawan uplink frame which is not
         // a proprietary frame
         Self::parse_frame(Direction::Uplink, self.payload())
-            .map(|frame| !matches!(frame, PHYPayloadFrame::Proprietary(_)))
+            .map(|frame| match frame {
+                PHYPayloadFrame::Proprietary(_) => false,
+                PHYPayloadFrame::JoinAccept(_) => false,
+                _ => true,
+            })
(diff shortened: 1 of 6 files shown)


src/packet.rs
```diff
@@ -113,7 +113,23 @@ impl PacketUp {
 
     pub fn is_potential_beacon(&self) -> bool {
         Self::parse_header(self.payload())
-            .map(|header| header.mtype() == lorawan::MType::Proprietary)
+            .map(|header| {
+                header.mtype() == lorawan::MType::Proprietary
+                    && self.payload().len() >= beacon::BEACON_PAYLOAD_SIZE
```
mawdegroot, line 118: `self.payload()` still includes the header byte so it should be `self.payload().len() - 1 >= beacon::BEACON_PAYLOAD_SIZE`

madninja: I know, but `>=` captures that case and allows for larger beacon payloads :-)

mawdegroot: True, I was just surprised by the off by one in my test case. It should sufficiently filter proprietary packets this way.

mawdegroot (approved):

mawdegroot (approved):

src/packet.rs
madninja: Fixed here [e64be24](https://github.com/helium/gateway-rs/pull/437/commits/e64be247d842c242bd7c8d32191e80a8078c916a)

PR #438, opened 2023-07-30
Session key for packet_router

madninja:
* Requires https://github.com/helium/proto/pull/359 to merge and local Cargo.toml updated
* Expects router to be upgraded to support session keys. 

NOTE: On a connect all pending uplink packets are sent but only after the session key is established to avoid a large rush of signatures before the session key ie active. Two other alternatives are to (1) add a delay before sending queued packets with whatever key is available (gateway or session) or (2) send all waiting packets and then deal with session key initialization when it arrives after that

### DIFF
--- a/src/packet_router/mod.rs
+++ b/src/packet_router/mod.rs
@@ -258,7 +258,6 @@ pub async fn mk_uplink(
     use std::ops::Deref;
     let mut uplink: PacketRouterPacketUpV1 = packet.deref().into();
     uplink.hold_time = packet.hold_time().as_millis() as u64;
-    uplink.gateway = keypair.public_key().into();
     uplink.signature = sign(keypair, uplink.encode_to_vec()).await?;
     let envelope = envelope_up_v1::Data::Packet(uplink);
     Ok(envelope)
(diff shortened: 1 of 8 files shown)


src/packet_router/mod.rs
```diff
@@ -192,22 +237,45 @@ impl PacketRouter {
...
+    let session_keypair = Arc::new(mk_session_keypair());
+    let session_key = session_keypair.public_key();
+
+    let mut session_init = PacketRouterSessionInitV1 {
+        gateway: keypair.public_key().into(),
+        session_key: session_key.into(),
+        nonce: offer.nonce.clone(),
+        signature: vec![],
+    };
+    session_init.signature = sign(session_keypair.clone(), session_init.encode_to_vec()).await?;
```
ke6jjj, line 278: I'm not sure this is being signed by the right key. It should be signed by the gateway's registered keypair (e.g. the ECC608), not by the session key itself.

ke6jjj (changes requested):

src/packet_router/mod.rs
madninja: Good catch... Fixed in [4767a67](https://github.com/helium/gateway-rs/pull/438/commits/4767a67d3148140d1482a722487f02f2ae264e05)

PR #452, opened 2023-08-17
Update keep-alive timer to prevent router from closing connections

joecryptotoo:
updated TCP keep-alive from 300 seconds to 30 seconds to prevent local router for closing the connection


madninja:
Closing.. let us know if this needs to be configurable. It does not seem to be a common occurrence

Issue #457, opened 2023-09-09
Remove chain fee from `add` command.


jthiller:
The `add` command currently returns a 65000 dc fee in addition to listing the 10M DC onboarding fee. This can be removed in favor of Solana fees.

https://github.com/helium/gateway-rs/blob/de873d8f6a529e9341123957047555b02fb0239f/src/cmd/add.rs#L38
https://github.com/helium/gateway-rs/blob/de873d8f6a529e9341123957047555b02fb0239f/README.md?plain=1#L367

PR #461, opened 2023-09-22
Support solana addresses for owner and payer

### LINKED ISSUE #460: Support Solana accounts in `add` command.
Currently user must look up their legacy L1 address in order to onboard a data-only Hotspot.

Should allow users to onboard using their Solana address. 

Currently lookup looks like,
```sh
./helium-wallet info
```
```json
{
  "address": {
    "helium": "<Helium Address>",
    "solana": "<Solana Address>"
  },
  "pwhash": "Argon2id13",
  "sharded": false
}
```

Add currently looks like,
```sh
./helium_gateway add --owner <helium_OWNER> --payer <helium_PAYER>
```

Solana OWNER or PAYER returns an error

madninja:
For legacy usage support this supports both helium public keys as well
as solana addresses.

Fixes #460


madninja:
@jthiller so the new output from the add cli here will now not display the fees since they're all determined by the network.. It'll also display owner and payer as Solana keys instead of the helium equivalent regardless of whether the user passed in legacy helium owner/payer or solana keys. 

eg. output:

```json
{
  "address": "helium_key",
  "mode": "dataonly",
  "owner": "solana_key",
  "payer": "solana_key",
  "txn": "xxx"
}
```

jthiller:
Lovely! I'll update the docs

jthiller:
closes #457

PR #462, opened 2023-09-28
Add TPM raw handle support

isergieienkov:


src/keypair.rs
```diff
@@ -103,13 +103,20 @@ impl FromStr for Keypair {
...
+                let key_identifier = &url.path()[1..];
+                let key_access = url.host().unwrap();
 
-                let keypair = tpm::Keypair::from_key_path(network, path)
-                    .map(helium_crypto::Keypair::from)
-                    .map_err(|err| {
-                        uri_error!("could not load tpm keypair on path {path}: {err:?}")
-                    })?;
+                let keypair = match key_access {
+                    "esys" => tpm::KeypairHandle::from_key_handle(
```
JayKickliter, line 110: What led you to add this match if you're only matching on `esys`? Is there another `key_access` that isn't being handled?

madninja (approved):

src/keypair.rs
isergieienkov: ah. there were 3 different options before I removed old FAPI

PR #474, opened 2023-11-21
adjust first beacon offset

madninja:
This adjusts the beaceon time to select a beacon time in the full first window, and moves it to the next window if in the past

### DIFF
--- a/src/beaconer.rs
+++ b/src/beaconer.rs
@@ -342,12 +342,17 @@ fn mk_next_beacon_time(
                 next_segment + random_duration(interval)
             } else {
                 // previous segment: Pick a time in the remainder of this
-                // segment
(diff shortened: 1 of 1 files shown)


mawdegroot (approved):
So, breaking down the logic here (for posterity):

When no beacon time exists, pick a random beacon time in the _current_ window and recursively call the `fun` again. Then
1. the beacon time is in the future and that time will be maintained. This can lead to an `irregular interval` beacon if the hotspot already beaconed in the current window.
2. the beacon time is in the past and a new beacon will be schedules in the _next_ window. This will not lead to an `irregular interval` beacon but may lead to a missed beacon

In both cases the normal schedule will resume after this logic.

PR #475, opened 2023-12-06
Data packet acks

madninja:
This adds a feature to allow for data packet acks to be sent back from the packet router to the gateway once they have been received. This feature can be used for deployments where backhaul is not reliable and packets are high value and not redundant enough to handle backhaul disconnects.


src/packet_router/mod.rs
```diff
@@ -55,26 +56,38 @@ pub struct PacketRouter {
...
-        let service =
-            PacketRouterService::new(router_settings.uri.clone(), settings.keypair.clone());
+        let service = PacketRouterService::new(
+            router_settings.uri.clone(),
+            router_settings.ack_timeout(),
+            settings.keypair.clone(),
+        );
         let store = MessageCache::new(router_settings.queue);
         let reconnect = Reconnect::default();
+        let ack_timer = AckTimer::new(router_settings.ack_timeout());
```
Vagabond, line 83: do we want to add some leeway here? the router currently sends exactly on this interval, so maybe add 1 or 2 seconds?

src/service/mod.rs
```diff
@@ -59,3 +59,35 @@ impl Reconnect {
...
+            next_time: Instant::now() + timeout,
+            timeout,
+        }
+    }
+
+    pub async fn wait(&self) {
+        if self.next_time >= Instant::now() {
+            time::sleep_until(self.next_time).await
+        } else {
+            std::future::pending().await
```
JayKickliter, line 80: This is surprising to see. Is it necessary? The reconnect timer has only the first branch.

JayKickliter: Oh, is it to briefly give the scheduler a chance to do something else?

src/service/mod.rs
```diff
@@ -59,3 +59,35 @@ impl Reconnect {
...
+
+pub struct AckTimer {
+    next_time: Instant,
+    timeout: Duration,
+}
+
+impl AckTimer {
+    pub fn new(timeout: Duration) -> Self {
+        Self {
+            next_time: Instant::now() + timeout,
```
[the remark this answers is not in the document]
JayKickliter: > What happens when timeout is 0 here?

`AckTimer::wait` will complete immediately the first time its called.

src/packet_router/mod.rs
madninja: addressed in [697c526](https://github.com/helium/gateway-rs/pull/475/commits/697c5261dfdbbc597123224b0a529b06d7c51031)

src/service/mod.rs
madninja: > This is surprising to see. Is it necessary? The reconnect timer has only the first branch.

It's there in case the ack_timeout is not needed at all (i.e. 0).. having a next_time in the past will just cause the sleep to fire immediately. Using a pending will just have it be a no-op await in the packet router Tokio select

PR #480, opened 2024-04-13
fix display of PacketUp

lthiery:
I believe the display implementation was copied from [semtech-udp](https://github.com/helium/semtech-udp/blob/eccaa208126aca962b6b349320fafe6712a6ec79/src/packet/push_data.rs#L219-L237).

However, frequency in that context is unfortunately a f64, but here it is a usize MHz. This PR displays it in human friendly MHz without using additional dependencies.


lthiery:
closed in favor of #481

PR #481, opened 2024-04-17
fix display of PacketUp

lthiery:


lthiery:
Anything wrong with this? FWIW, it's solving a display bug where MHz units are used but the output is Hz.

madninja:
> Anything wrong with this? FWIW, it's solving a display bug where MHz units are used but the output is Hz.

Been a while! Iirc I was considering trying to do without rust-decimal to avoid the size overhead it carries. Os it already used in one of the deps?

lthiery:
Looks in the workspace and Cargo.lock already but unsure if it's linking in anything new. How big could it be anyway?

lthiery:
I run
```
git checkout main && cargo clean && cargo build --release && ls -lh target/release/helium_gateway
````
which gives
```
-rwxrwxr-x 2 louis louis 3.4M Apr  2 20:14 target/release/helium_gateway
```
And then I run: 
```
git checkout lthiery/packet-up-decimal  && cargo update time && cargo clean && cargo build --release && ls -lh target/release/helium_gateway
```
which gives
```
-rwxrwxr-x 2 louis louis 3.4M Apr  2 20:17 target/release/helium_gateway
```

Let's fix the display bug

madninja (approved):

PR #489, opened 2025-04-03
updates such that gateway-rs builds as lib

lthiery:
gateway-rs is not up to date with latest helium-proto and the declaration on "main" is inaccurate

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ members = ["lorawan"]
 byteorder = "1"
 serde = { version = "1", features = ["rc", "derive"] }
 rust_decimal = { version = "1", features = ["serde-with-float"] }
-helium-proto = { git = "https://github.com/helium/proto", rev = "d94ed4b", features = [
+helium-proto = { git = "https://github.com/helium/proto", branch = "master", features = [
     "services",
 ] }
 rand = "0.8"
@@ -70,7 +70,7 @@ signature = { version = "1", features = ["std"] }
 async-trait = "0"
 angry-purple-tiger = "0"
 lorawan = { package = "lorawan", path = "lorawan" }
-beacon = { git = "https://github.com/helium/proto", rev = "d94ed4b" }
+beacon = { git = "https://github.com/helium/proto", branch = "master" }
 exponential-backoff = { git = "https://github.com/yoshuawuyts/exponential-backoff", branch = "master" }
 semtech-udp = { version = ">=0.12", default-features = false, features = [
     "server",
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -18,7 +18,7 @@ members = ["lorawan"]
 byteorder = "1"
 serde = { version = "1", features = ["rc", "derive"] }
 rust_decimal = { version = "1", features = ["serde-with-float"] }
-helium-proto = { git = "https://github.com/helium/proto", branch = "master", features = [
+helium-proto = { git = "https://github.com/helium/proto", rev = "d94ed4b", features = [
```
madninja, line 21: Sorry what happens on master that breaks things here? I see another fix went in in a different downstream crate to work around things. May be easier to just adjust code?

lthiery:
```
error[E0004]: non-exhaustive patterns: `Ok(helium_proto::services::router::envelope_down_v1::Data::PacketAck(_))` not covered
```

lthiery: looks like a new message exists

madninja: we should be able to handle that though..

madninja (commented):
The build issue isn't on the clippy side, it's the cross toolchian not supporting the newer rust version yet that bumps the lock version apparently?

lthiery:
> The build issue isn't on the clippy side, it's the cross toolchian not supporting the newer rust version yet that bumps the lock version apparently?

Yea I was seeing if I could bump the toolchain version but too much of a hassle

Cargo.toml
lthiery: Yea i'm just not sure what to do with it. I'm just trying to get this usable as a lib at the moment

madninja: I believe it was added to HPR as part of https://github.com/helium/gateway-rs/pull/475

lthiery:
Shall we just deprecate the quay step and leave its recovery as an open issue?

Cargo.toml
madninja: I still believe the PR above should be fixed and merged up or this PR fixed to handle the case

madninja:
> Shall we just deprecate the quay step and leave its recovery as an open issue?

Nope.. looking at it.. perhaps address the other comments?

lthiery:
> > Shall we just deprecate the quay step and leave its recovery as an open issue?
> 
> 
> 
> Nope.. looking at it.. perhaps address the other comments?

I addressed all comments and was awaiting response. Things were stalled on quay

Cargo.toml
lthiery: I would think be able to build at all would be a good enough PR on its own.

Updating is out of scope of what I can do at the moment

madninja (approved):

madninja (changes requested):
Looks like it missed a spot

Cargo.toml
lthiery: I implemented a warning but there is also packet_ack_interval in PacketRouterRegisterV1

madninja: that's ok, as long as it's not set I believe HPR will not issue regular acks

madninja (approved):

PR #491, opened 2026-01-13
Update dependencies for tonic 0.14 and helium-proto compatibility

lthiery:
- Update tonic from 0.x to 0.14
- Update http from * to 1.3
- Update http-serde from 1 to 2
- Update tokio to 1.48
- Migrate from helium_proto::services::{Channel, Endpoint} to tonic::transport::{Channel, Endpoint}
- Update error types from helium_proto::services::Error to tonic::transport::Error
- Add helium_proto, helium_crypto, semtech_udp re-exports
- Use helium-crypto fork with signature std feature fix

Depends on: https://github.com/lthiery/helium-crypto-rs/main for now


madninja (commented):
You actually have demand for mips? Iirc there were basically no hotspots that use it?

PR #493, opened 2026-04-03
Fix off-by-one error in message cache size limit

prjanitor:
The cache was allowing max_messages + 1 items before trimming because the boundary check used `>` instead of `>=`. This caused the cache to exceed its intended maximum size by one element.

Changed the comparison operator from `>` to `>=` in both:
- `push_back()` method
- `push_front()` method

This ensures the cache is trimmed when it reaches the maximum size, not when it exceeds it.

---

*This PR was generated by [PRJanitor](https://prjanitor.com) — an automated tool that finds and fixes small bugs in open-source projects.*

We respect your contribution guidelines — if your project doesn't accept bot PRs, we won't send more. You can also add a `.github/prjanitor.yml` file with `enabled: false` to opt out explicitly.

### DIFF
--- a/src/message_cache.rs
+++ b/src/message_cache.rs
@@ -51,7 +51,7 @@ impl<T: PartialEq> MessageCache<T> {
     /// (first) message in the cache to be dropped.
     pub fn push_back(&mut self, message: T, received: Instant) -> Option<CacheMessage<T>> {
         self.cache.push_back(CacheMessage::new(message, received));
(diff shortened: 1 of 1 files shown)


prjanitor:
Closing this PR -- after a closer review, we believe this change isn't correct. Apologies for the noise; we're still ironing out some kinks in our automated flow. Thanks for your patience!

